PR #2, opened 2016-02-13
RH: Change lonelycode to TykTechnologies after the repo moved

richard-hulm:
It seems I couldn't build this locally because it refers to lonelycode rather than TykTechnologies

Changing this made it work OK


lonelycode:
Good spot, thanks!

PR #4, opened 2016-02-14
RH: Added ability to put all data into a new index every day 

richard-hulm:
Rather than one static index

In ES, this enables easy backups to be taken of individual days worth of logs, and for index settings changing at midnight without having to reindex all existing data

Also removed the explicit "CreateIndex" call as by default when you insert a document into an index for the first time it creates the index in ES. Could add this back in as an option later

Branched off from develop this time? Which would you prefer I worked from / merged into?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## VX.X
 
 - Initial connector which outputs all analytics into a single elasticsearch index
+- Enabled the ES connecter to save all output to a rolling index rather than a static index
 
 ## v0.1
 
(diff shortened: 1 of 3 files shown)


lonelycode:
Amazing - thanks :-)

lonelycode:
We work from develop usually, have merged all your other PRs back into dev so master and develop are in sync. If possible, work against develop as that's where most of the new stuff is, but we try to keep master stable. 

In future master will be sacrosanct (and version tagged), with develop being the primary unstable branch. 

Keeps it clean for people wanting to pull and compile themselves.

Really appreciate these - really cool stuff :-)

PR #7, opened 2016-04-13
Add Segment analytics pump

JosephEarl:
Adds a pump for https://segment.com

It would be nice for an analytics pump to be able to identify keys - i.e. when a key is created the pump gets the chance to record the email/meta data associated with that key.


lonelycode:
Can you make this PR against develop? We've added more pumps since (statsd and influx), will be easier to merge into develop rather than master

lonelycode:
(this is very cool btw)

JosephEarl:
Will do

JosephEarl:
Done https://github.com/TykTechnologies/tyk-pump/pull/8

PR #8, opened 2016-04-13
Add segment analytics pump

JosephEarl:
Adds pump for Segment.com

### DIFF
--- a/pumps/segment.go
+++ b/pumps/segment.go
@@ -1,90 +1,95 @@
 package pumps
 
 import (
-	"fmt"
-	"encoding/json"
-	"errors"
-	"github.com/Sirupsen/logrus"
-	segment "github.com/Segment/analytics-go"
+  "encoding/json"
+  "github.com/Sirupsen/logrus"
+  "github.com/TykTechnologies/tyk-pump/analytics"
+  "github.com/mitchellh/mapstructure"
+  segment "github.com/segmentio/analytics-go"
 )
 
 type SegmentPump struct {
-	segmentClient *segment.Client
-	segmentConf   *SegmentConf
+  segmentClient *segment.Client
+  segmentConf   *SegmentConf
 }
 
 var segmentPrefix string = "segment-pump"
 
 type SegmentConf struct {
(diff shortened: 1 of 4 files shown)


pumps/segment.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (s *SegmentPump) WriteData(data []interface{}) error {
+	log.WithFields(logrus.Fields{
+		"prefix": segmentPrefix,
+	}).Info("Writing ", len(data), " records")
+
+	for i, v := range data {
+		analyticsRecord := AnalyticsRecord{}
+		err := msgpack.Unmarshal(v.([]byte), &analyticsRecord)
```
lonelycode: Does this actually work?? You shouldn't need to unmarshal the data using msgpack, it's already unmarshalled when it hits this handler. There is a special case for uptime data, which issn't handled in other pumps. See the [influx handler](https://github.com/TykTechnologies/tyk-pump/blob/develop/pumps/influx.go#L92) for proof ;-)

JosephEarl: I copied this over from some code I had against the 1.9 pump before it was separated out, so may not be needed anymore.

You are saying I can just do 
```
analyticsRecord := v.(AnalyticsRecord)
```
?

lonelycode: Yup, no need to try and decode (that would generate an error I think).

JosephEarl:
Removed the JSON decoding

lonelycode:
Stupid question, but I gotta ask - but have you successfully used this with segment?

JosephEarl:
I'm still doing some testing with Segment so I'd wait a bit before merging just to confirm it's OK.

JosephEarl:
Updated and added a test (ignored currently - but you can enter your own Segment key and remove the Skip to try). Here's a screenshot of an event in the Segment debugger:

segmenttrackhit

lonelycode:
Amazing! Will merge, should be available in the nightlies in about an hour (will publish URLs in the community forum soon)

JosephEarl:
Great!

PR #11, opened 2016-06-16
Features/graylog

duythinht:
I just add support for graylogs. Tested working with graylog2 version >= 1.3.4
@lonelycode Please take a reviews

### DIFF
--- a/pumps/graylog.go
+++ b/pumps/graylog.go
@@ -7,7 +7,6 @@ import (
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/mitchellh/mapstructure"
 	"github.com/robertkowalski/graylog-golang"
-	"strconv"
 )
 
 type GraylogPump struct {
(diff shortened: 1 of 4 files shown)


pumps/graylog.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"prefix": graylogPrefix,
+			}).Fatal(err)
+		}
+
+		gelfData := map[string]interface{}{
+			//"version": "1.1",
+			"host":      "tyk-pumps",
+			"timestamp": record.TimeStamp.Unix(),
```
richard-hulm: Do you not want to use the timestamp from the analytics data rather than the current timestamp?

duythinht: It was using analytics timestamp from the record

richard-hulm: .......Completely misread that sorry!

lonelycode:
LGTM :-)

PR #14, opened 2016-07-12
Some changes to how fields are mapped for Elasticsearch output

richard-hulm:
Sorry - got it slightly wrong before, but added a missed field and updated the keys for a couple of the fields.


lonelycode:
Snap! :-)

PR #18, opened 2016-12-20
Elasticsearch Extended Stats Mode

catdevnull:
Adds an "extended_stats" flag which changes the behavior of the Elasticsearch pump to include the following additional fields in the Elasticsearch payload:

* Raw Request
* Raw Response
* User Agent

Defaults to false if not specified which should ensure backward compatibility with existing users. Depending on raw request or response sizes could impact on performance which is why this has been implemented as an optional flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Create a `pump.conf` file:
                 "elasticsearch_url": "localhost:9200",
                 "enable_sniffing": false,
                 "document_type": "tyk_analytics",
-                "rolling_index": false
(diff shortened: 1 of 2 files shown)


catdevnull:
Compile fail on semaphore. Will review and reopen a new pull request.

lonelycode:
Ah semaphore may always fail disregard that the integration is currently broken :-/

catdevnull:
It's alright I'd made a syntax error in the code. Not used to Go Lang syntax!

PR #19, opened 2016-12-20
Elasticsearch Extended Stats Mode

catdevnull:
Adds an "extended_stats" flag which changes the behavior of the Elasticsearch pump to include the following additional fields in the Elasticsearch payload:

* Raw Request
* Raw Response
* User Agent

Defaults to false if not specified which should ensure backward compatibility with existing users. Depending on raw request or response sizes could impact on performance which is why this has been implemented as an optional flag.

I've also added very brief compile instructions to the README to assist those in the future who are new to Go compilation!


lonelycode:
Cool, could you target `develop` though? Master is for stable release only now.

catdevnull:
@lonelycode done!

Issue #20, opened 2017-01-12
Upgrade to Elastic v5


kohrVid:
The current version of the Tyk Pump only supports ElasticSearch version 3. Can `pump/elasticsearch.go` please be updated to use version 5 of the Elastic package ( gopkg.in/olivere/elastic.v5) which supports ElasticSearch v5?

PR #22, opened 2017-01-19
Upgrade to support Elasticsearch 5

miles-b:
A simple change, just using version 5 of the Go Elastic package (`gopkg.in/olivere/elastic.v5`), as suggested in #20. The full list of breaking changes between 3 and 5 is [here](https://github.com/olivere/elastic/blob/release-branch.v5/CHANGELOG-5.0.md#enforce-contextcontext-in-performrequest-and-do), not much that affects us, other than the `Do()` method now requires a `context.context`. I have tested this with Elasticsearch 5.0.1. 
Will break compatibility with older versions of Elasticsearch; I did experiment with a [more complex change that would support both versions](https://github.com/TykTechnologies/tyk-pump/compare/develop...miles-b:feature/elasticsearch-5), but in the end I decided its probably better to keep it simple

### DIFF
--- a/pumps/elasticsearch.go
+++ b/pumps/elasticsearch.go
@@ -4,8 +4,9 @@ import (
 	"github.com/TykTechnologies/logrus"
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/mitchellh/mapstructure"
-	"gopkg.in/olivere/elastic.v3"
+	"gopkg.in/olivere/elastic.v5"
 	"time"
+	"context"
 )
 
 type ElasticsearchPump struct {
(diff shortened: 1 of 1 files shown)


mvdan:
@lonelycode is dropping earlier ES versions acceptable? If so, this is definitely better than all the code linked to support both versions.

lonelycode:
@mvdan Not hugely happy with losing support for older versions - the version should be configurable IMO.

An easy way to achieve this would be to create a new pump for ESv5 (basically this change but under a new name, e.g. 

`var elasticsearchPrefix string = "elasticsearchv5-pump"`

and ensure it can be [loaded](https://github.com/TykTechnologies/tyk-pump/blob/master/pumps/init.go#L8), then existing functionality is not lost (we do not want to force an upgrade on our users because of this component since many users will also have ES running for other services).

miles-b:
@lonelycode Sure - I created another version that is backward compatible at https://github.com/TykTechnologies/tyk-pump/compare/develop...miles-b:feature/elasticsearch-5 - would that be better then? I did consider creating a new pump, but the code was so similar, I had a go at integrating both versions in the same pump. It uses a new "version" field in pump.conf that takes either "3" or "5" (default if omitted is 5, but that can be changed to match current behaviour)

lonelycode:
@miles-b The backwards compatible version looks good to me, I'd prefer if it defaulted to 3 and required a version statement for 5, that way we don;t break anyone else's systems on release :-) 

Want to submit it as a PR?

miles-b:
@lonelycode sorry for the delay, been very busy recently. I've made the changes you suggested and created #23 , so I'll close this now :)

PR #29, opened 2017-03-15
add alpine docker file

jtuthehien:
Add option to build docker image, alpine based.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.5
+WORKDIR /tyk-pump
+COPY ./tyk-pump/pump.example.conf /tyk-pump/pump.conf
+COPY ./dist/tyk-pump /tyk-pump/tyk-pump
+CMD ["./tyk-pump","--c=/tyk-pump/pump.conf"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+package=github.com/TykTechnologies/tyk-pump
+distname=tyk-pump
+goversion=1.8.0
+
+build:
+	docker run -v `pwd`:/go/src/$(package) golang:$(goversion)-alpine go build -o /go/src/$(package)/dist/$(distname) -v $(package)
+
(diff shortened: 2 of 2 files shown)


mvdan:
Thanks for the patch!

The repo you want is https://github.com/TykTechnologies/tyk-pump-docker-pub/tree/version/0.4 (notice the branch). I realise this is confusing, but it's where the current dockerfile is.

Also, if what you're after is to reduce the image size, you can do most of that without changing the base distro, e.g. https://github.com/TykTechnologies/tyk-dashboard-docker/pull/4.

jtuthehien:
I'm aware of the docker repo, and it's confusing since you tie the binary into ubuntu package which makes it pretty hard to do anything else.
What it really needs is just the single binary and pretty can be run on all linux distro. 
Running everything in one RUN command does reduce the image size, but still capped by gigantic ubuntu based. Alpine based is just 10MB

lonelycode:
Thanks for the PR! I'm afraid we're going to have to reject it though :-(

We're not going to start putting Dockerfiles with the repository, because we only guarantee that release branches will build and be stable and it would add **yet-another-thing-that-needs-managing** as part of cutting code for the team, I'd prefer to have the dockerfiles independent of the repo so that they can be versioned appropriately and managed independently without touching the base project. 

Also, I'd prefer not to take a stance on "which docker base image is best", so this should live in a separate project.

Completely understand that we publish on an ubuntu base image and that doesn't work for you, but I would strongly suggest that you base your images on stable releases that we have published (with all supporting files, configs and any other things that have gone into making sure it runs and is stable. This approach would fail with things like the dashboard and the gateway.

If needs must, extract the binary from the deb file using the appropriate tools and dockerize that, the binary is the same for all linux variants we support, only the supporting scripts and packaging/signing are different. At least that way you will also get the correct file structure for the app.

I'd propose we create a new project for these kinds of things, then we can have multiple docker files for whatever base image users want, and it can be maintained by the community?

PR #30, opened 2017-03-17
Demo data

lonelycode:
Generates demo data for any specified organisation, usage:

```
./tyk-pump -c ./pump.conf -d {org-id}
```

OR, for containers set the env var:

```
export TYK_PMP_BUILDDEMODATA=589a2058c2e9f9018bad03aa
```

Will generate 30 days or so of demo data for a set of APIs, popularting data as necessary. Only GeoData is not included

### DIFF
--- a/analytics/demo/demo.go
+++ b/analytics/demo/demo.go
@@ -1,0 +1,174 @@
+package demo
+
+import (
+	"time"
+	"math/rand"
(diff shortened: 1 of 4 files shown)


lghiur:
@lonelycode, thanks for this, it works great! One small correction, it should be "-demo" instead of "-d", @mvdan helped with that.

PR #33, opened 2017-04-03
Get rid of the infinite recursive tail call in StartPurgeLoop

### LINKED ISSUE #27: Potential memory leak in the pump over time
Tyk pump has a potential memory leak over time, it is confirmed to with the standard mongo pumps, but it exacerbated by the elasticsearch pump

Behaviour was seen with 10s purge delays

mvdan:
The loop needs to run forever. Do that in an actual loop instead of
using recursive tail calls. That will surely lead to a growing call
stack, eventually leading to runtime panics.

Fixes #27.

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,48 +103,46 @@ func initialisePumps() {
 
 }
 
-func StartPurgeLoop(nextCount int) {
-	job := instrument.NewJob("PumpRecordsPurge")
+func StartPurgeLoop(secInterval int) {
+	for range time.Tick(time.Duration(secInterval) * time.Second) {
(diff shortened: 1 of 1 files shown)


mvdan:
Really easy to review ignoring whitespace: https://github.com/TykTechnologies/tyk-pump/pull/33/files?w=1

mvdan:
Also note that the only behaviour change is that the loop now runs at time 0 too, which it didn't before. I assume that's fine. If not, I can make the logic slightly more complicated to accomodate.

lonelycode:
*looks

Issue #16, opened 2016-10-17
No reconnecting to database.


eloyucu:
Hi everyone.

I was getting a error with the stats that pump must handle. The problem is that the stats wasn't being shown. The line that make reference in the syslog: 

> level=error msg="Problem inserting to mongo collection: EOF" 

It seems that tyk-pump cant reconnect to the mongo database. It seems that this [link](https://community.tyk.io/t/mongo-db-re-connection/158/3) is not totally correct (I mean, someone says that the problem is fixed... but no). 

I stop and restart mongo and I obtain this error, and then if I restart pump... it works again.

lonelycode:
Which version are you running?

eloyucu:
Ooopsss!! Sorry, I forgot put the versions...

tyk-dashboard -> 1.1.0.0
tyk-gateway     -> 2.1.0.2
tyk-pump         -> 0.2.0.0

Thanks.

Preravana:
Hi, In documentation at pump.conf file it is provided host and hosts what does it mean actually. what should be the value of host and hosts? I am using Docker installation.

eloyucu:
Hi @Preravana 

I think you can open a new issue for your question... because 1) it is a old issue 2) it is not related with yours.

Nice to meet you.

PR #42, opened 2017-08-04
Allow mongo connection errors to fail

jspc:
Last night we suffered an outage where pump could no longer talk to mongo and redis filled up, causing things like gateway enroll/re-enrolls and key operations to fail.

This looks related to https://github.com/TykTechnologies/tyk-pump/issues/16

This was due to, or at least could have been mitigated by, `mgo.DialWithInfo()` not passing a `*mgo.DialInfo` with a timeout or the `FailFast` flag being set- the connection churned and churned and churned.

This PR proposes an optional config key: `mongo_connect_fail_fast`. This is used to both fail the connection fast, and to kill the process should this occur.

Killing the process allows monitoring systems to pick the defect up. In our case this would have solved the connection problem we had. For other's it'll help alert people to problems.

*Note:* the redis client also keeps churning away silently- why?

### DIFF
--- a/pump.example.conf
+++ b/pump.example.conf
@@ -1,5 +1,5 @@
 {
-	"analytics_storage_type": "redis",
+    "analytics_storage_type": "redis",
     "analytics_storage_config": {
         "type": "redis",
         "host": "localhost",
@@ -18,7 +18,8 @@
             "name": "mongo",
             "meta": {
                 "collection_name": "tyk_analytics",
-                "mongo_url": "mongodb://localhost/tyk_analytics"
+                "mongo_url": "mongodb://localhost/tyk_analytics",
+                "mongo_connect_fail_fast": true
             }
         }
     },
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -2,6 +2,7 @@ package pumps
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net"
 	"strings"
 	"time"
@@ -32,13 +33,16 @@ type MongoConf struct {
 	MongoSSLInsecureSkipVerify bool   `mapstructure:"mongo_ssl_insecure_skip_verify"`
 	MaxInsertBatchSizeBytes    int    `mapstructure:"max_insert_batch_size_bytes"`
 	MaxDocumentSizeBytes       int    `mapstructure:"max_document_size_bytes"`
+	FailFast                   bool   `mapstructure:"mongo_connect_fail_fast"`
 }
 
-func mongoDialInfo(mongoURL string, useSSL bool, SSLInsecureSkipVerify bool) (dialInfo *mgo.DialInfo, err error) {
+func mongoDialInfo(mongoURL string, useSSL, SSLInsecureSkipVerify, failFast bool) (dialInfo *mgo.DialInfo, err error) {
 	dialInfo, err = mgo.ParseURL(mongoURL)
 	if err != nil {
 		return dialInfo, err
 	}
+
+	dialInfo.FailFast = failFast
 
 	if useSSL {
 		dialInfo.DialServer = func(addr *mgo.ServerAddr) (net.Conn, error) {
@@ -107,7 +111,7 @@ func (m *MongoPump) connect() {
 	var err error
 	var dialInfo *mgo.DialInfo
 
-	dialInfo, err = mongoDialInfo(m.dbConf.MongoURL, m.dbConf.MongoUseSSL, m.dbConf.MongoSSLInsecureSkipVerify)
+	dialInfo, err = mongoDialInfo(m.dbConf.MongoURL, m.dbConf.MongoUseSSL, m.dbConf.MongoSSLInsecureSkipVerify, m.dbConf.FailFast)
 	if err != nil {
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
@@ -115,10 +119,20 @@ func (m *MongoPump) connect() {
 	}
 
 	m.dbSession, err = mgo.DialWithInfo(dialInfo)
+
 	if err != nil {
+		var errorStr = fmt.Sprintf("Mongo connection failed: %q", err.Error())
+
+		if m.dbConf.FailFast {
+			log.WithFields(logrus.Fields{
+				"prefix": mongoPrefix,
+			}).Panic(errorStr)
+		}
+
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
-		}).Error("Mongo connection failed:", err)
+		}).Error(errorStr)
+
 		time.Sleep(5)
 		m.connect()
 	}
--- a/pumps/mongo_aggregate.go
+++ b/pumps/mongo_aggregate.go
@@ -3,6 +3,7 @@ package pumps
 import (
 	b64 "encoding/base64"
 	"errors"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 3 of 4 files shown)


lonelycode:
Thanks for this - in general we try not to kill the process, since then we must assume that there is a monitoring system in place, and that the OS or the process monitor will restart the application. 

We do test connection failures and haven't seen this happen in a very long time (nor have we had other reports), so this scenario might be new.

Could you let us know the exact nature of the failure? It might help us identify a strategy.

buger:
Is there anything in pump logs, related to mongo timeout issues? Why did you decide that it was mongo?

Looking at the driver documentation `Dial` function:
> Dial will timeout after 10 seconds if a server isn't reached. The returned session will timeout operations after one minute by default if servers aren't available

And in addition, if the error happens we have additional retry logic, which should handle reconnect after a 5-second timeout.

`FailFast` doc says:

> FailFast will cause connection and query attempts to fail faster when the server is unavailable, instead of retrying until the configured timeout period. Note that an unavailable server may silently drop packets instead of rejecting them, in which case it's impossible to distinguish it from a slow server, so the timeout stays relevant.

So it disables internal driver restarts, by our restart should still work, and timeout still applied. 

I can see such situation when analytics data does not get fetched from Redis, if mongo is down for some reason, and `StartPurgeLoop` wait for it to connect. So if mongo was down and go up after some time, period of time when Pump will not work is: `<downtime> + max(1 minute, internal timeout) + max(5 second, tyk reconnect timeout)`

Regarding what can be done better, in this case, is reading info from Redis in any case, writing to pumps asynchronously, and if write queue becomes too large, drop analytics data. This will protect over Redis filling up if mongo is down. 

Also as mentioned by @lonelycode, killing the process is not the best idea, and actually make no much sense, since restarts already handled by the driver. 

So, can you confirm that your mongo had downtime? Or for example, you had mongo cluster but did not specify all addresses to the pump, so it can choose in case of failover?

Thank you!

jspc:
I can definitely confirm our mongo died; we received alerts from mongo cloud, which we use to manage our cluster, that the cluster had no master was elected for the replica set holding tyk config- this was due to some mongo maintenance.

From our redis elasticache cluster we can see that we run out of memory to store new keys:

elasticache grafana monitoring

Which manifests in gateway logs such as

```
time="Jul 31 16:54:11" level=error msg="OOM command not allowed when used memory > 'maxmemory'."
time="Jul 31 16:54:11" level=error msg="Error trying to delete keys:"
```

Which mean that as autoscaling events on AWS kick in and the gateways talk back to the dashboard, the dash can't enrol them and requests fail, in some instances, and in others means redis drops connections and API Key lookups fail.

See: https://flood.io/v2mIHP0s

If you don't mind, @lonelycode and @buger, I'll respond to both of your points/ questions in this message:

> Thanks for this - in general we try not to kill the process, since then we must assume that there is a monitoring system in place, and that the OS or the process monitor will restart the application 

In this instance the key will always default to `false` when omitted due to how golang defaults booleans (I know you're aware of this: I wanted to be clear for other people coming into the PR)

In my instance: I do have these things and I want these things to spot when pump isn't doing what we expect it to do

> Could you let us know the exact nature of the failure? It might help us identify a strategy.

A cluster/replica set (there's only one RS on this cluster so for our case they're analogous but I don't know which of the situations could cause this)  rebalance failed which took down mongo: it had to recover from not having a defined master. While we received monitoring about this, we want to know all dependent services fail so we can see the net effect of a defect, and monitor to see all affected services recover. We couldn't do that here- we assumed that pump was fine.

> Looking at the driver documentation Dial function:

I couldn't/ still can't see that: the `mgo` doesn't set the timeout in https://github.com/go-mgo/mgo/blob/v2/session.go#L271 so I thought it'd stay at the default 0, meaning that it waits forever.

In fact; if I apply the following patch:

```patch
diff --git a/pumps/mongo.go b/pumps/mongo.go
index b956de7..b3c13d2 100644
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -2,6 +2,7 @@ package pumps

 import (
        "crypto/tls"
+       "fmt"
        "net"
        "strings"
        "time"
@@ -114,7 +115,9 @@ func (m *MongoPump) connect() {
                }).Panic("Mongo URL is invalid: ", err)
        }

+       fmt.Printf("%v : starting connection\n", time.Now())
        m.dbSession, err = mgo.DialWithInfo(dialInfo)
+       fmt.Printf("%v : connection returned with: %v, %v\n", time.Now(), m.dbSession, err)
        if err != nil {
                log.WithFields(logrus.Fields{
                        "prefix": mongoPrefix,
```

I can see that it's not dropping after ten seconds:

```bash
$ time ./tyk-pump -c pump.example.conf
[Aug  4 14:32:58]  INFO main: ## Tyk Analytics Pump, v0.4.2 ##
[Aug  4 14:32:58]  INFO mongo-pump: -- No max batch size set, defaulting to 10MB
[Aug  4 14:32:58]  INFO mongo-pump: -- No max document size set, defaulting to 10MB
2017-08-04 14:32:58.54240163 +0100 BST : starting connection
^C

real    1m26.319s
user    0m0.036s
sys     0m0.069s
```

> I can see such situation when analytics data does not get fetched from Redis, if mongo is down for some reason, and StartPurgeLoop wait for it to connect. So if mongo was down and go up after some time, period of time when Pump will not work is: <downtime> + max(1 minute, internal timeout) + max(5 second, tyk reconnect timeout)

Ditto above- my understanding from `mgo` is that it doesn't handle this.

> Also as mentioned by @lonelycode, killing the process is not the best idea, and actually make no much sense, since restarts already handled by the driver.

We want to see all dependant services die, so we can watch them recover. In our case we'd have seen the ASG tyk-pump sits in cycle instances and remain red/ in an alert state until the service returned.

> So, can you confirm that your mongo had downtime? Or for example, you had mongo cluster but did not specify all addresses to the pump, so it can choose in case of failover?

I can, or at least I can show symptoms:

mongo atlas alert


Ultimately: at the same time mongo went down our redis cluster started to fill up with analytics data. The cluster recovered once mongo had been fixed. This is where we assume mongo being down had an issue.

Had pump died, we'd have known pump was being affected and could have manually removed analytics to keep services which require redis up.

buger:
Make sense! 

Note that there is alternative to monitoring pump without killing it.
Pump provides statsd instrumentation which can be enabled by setting "TYK_INSTRUMENTATION=1" environment variable, and in config `statsd_connection_string` and `statsd_prefix` options. 

You should check for `PumpRecordsPurge` and `GCActivity` to get real-time information about the health of Pump.

Regarding `mgo` timeouts, I'm checking it, and if it is true (which is weird), we should explicitly set these timeouts in our code. 

Thank you!

jspc:
No worries- we're not using statsd, which is the only problem with using that. We could, though, set it locally and push somewhere but ideally I'd love to use the monitoring and alerting we currently have.

Beyond the fact you've got statsd and that you don't think killing a process is smart, do you have any objection to this PR where an opt-in kill the process config option works for other people?

(Just so I can figure out what my next steps are: if this is never going to get merged then I can put my effort elsewhere)

buger:
If we are going to do it properly, in my view we should do the following:

Killing the process is a bad practice and rarely used for server side software. I would prefer service to auto-heal. That's where the `FailFast` and explicit timeout comes in: I will investigate it in details and prepare the PR

In addition, Gateway should have a mechanism which prevents filling Redis with data in case of pump failure. It should monitor stats queue size, and drop analytics data. Limit of the queue should be configurable. 

Make sense?

buger:
Actually one more thing. We planning add HTTP health check endpoint to the pump, which can be used for monitoring and by load balancers. This health check endpoint will also check Redis and Mongo connectivity, and will return meaningful error if something wrong.

jspc:
Gateway makes sense, yeah.

From the process point: we expect our processes to fail fast so they're not running in a degraded state without us knowing. In our situation we specifically want this behaviour- we build out tooling around this with healthchecking, autoscaling, so on.

Killing the process is used quite often in server side software; especially when bootstrapping connections- I don't want things started/ running in a degraded state because that makes monitoring more difficult: we need to either consume another application's instrumentation or we need to monitor on log entries.

In any case- we'll find a solution which matches our configuration/ infrastructure. Hopefully the timeout stuff is useful!

PR #47, opened 2017-10-18
lint: cleaning mongo pump

MangoMM:


pumps/mongo.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/TykTechnologies/tyk-pump/analytics"
 )
 
-const tenMB int = 10485760
+const tenMB int = 10 << 20
```
joshblakeley: Is this just a readability preference?

pumps/mongo.go
```diff
@@ -115,159 +115,165 @@ func (m *MongoPump) connect() {
...
+
+	// TODO - Should this not bail after a while?
+	for err != nil {
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
 		}).Error("Mongo connection failed:", err)
+
 		time.Sleep(5)
-		m.connect()
+		m.dbSession, err = mgo.DialWithInfo(dialInfo)
```
joshblakeley: Is there a situation where you would prefer the recursive approach to iterative?

pumps/mongo.go
MangoMM: For me is less error prone, very fast and explicit - rather than an arbitrary number. If others have objections, will revert that bit.

pumps/mongo.go
MangoMM: Quicksort & traversing trees.

pumps/mongo.go
lonelycode: For the philistines like me, I don't know what this does, add a comment to explain?

PR #50, opened 2017-10-19
[WIP | PoC] Mongo capped collection

MangoMM:


pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
 	return nil
 }
 
+func (m *MongoPump) capCollection() {
+
+	var colName = m.dbConf.CollectionName
+	var colCapMaxSizeBytes = m.dbConf.CollectionCapMaxSizeBytes
```
mvdan: nit, but I'd prefer `foo := bar` for consistency

pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
...
+	}
+
+	if !exists {
+		if err := m.createCappedCollection(colName, colCapMaxSizeBytes); err != nil {
+			// TODO handle error
+
+			return
+		}
+
+		return
```
mvdan: I'd add a comment here, like `we're already capping the newly created collection`

pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
...
+	sess := m.dbSession.Copy()
+	defer sess.Close()
+
+	colNames, err := sess.DB("").CollectionNames()
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"prefix": mongoPrefix,
+		}).Error("Unable to get column names: ", err)
+
+		return false, err
```
mvdan: nit, but do `return false, fmt.Errorf("unable to get column names: %v", err)` and log in the caller

(it's generally frowned upon to both log and return an error, even if the old code does it quite often)

pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
...
+	}
+
+	// If colCapMaxSizeBytes == 0, uncap the collection and return
+	if colCapMaxSizeBytes == 0 {
+		m.uncapCollection(colName)
+
+		return
+	}
+
+	m.resizeCappedCollection(colName, colCapMaxSizeBytes)
```
lonelycode: Not sure we want to be resizing data on behalf of the user... it's destructive, let's avoid being destructive.

pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
...
+
+	// No need to create collection if MaxBytes are 0
+	if maxBytes == 0 {
+		return nil
+	}
+
+	return sess.DB("").C(name).Create(&mgo.CollectionInfo{Capped: true, MaxBytes: maxBytes})
+}
+
+func (m *MongoPump) uncapCollection(name string) {
```
lonelycode: Make sure this is possible in older versions of MongoDB, not everyone runs latest.

pumps/mongo.go
```diff
@@ -103,6 +107,115 @@ func (m *MongoPump) Init(config interface{}) error {
...
+	// If not capped, return
+
+	// If capped, do uncap logic:
+
+	// db.collection.copyTo("collection_temp")
+	// db.collection.drop()
+	// db.collection_temp.renameCollection("collection")
+}
+
+func (m *MongoPump) resizeCappedCollection(name string, maxBytes int) {
```
lonelycode: I don't think we should resize collections, see above.

MangoMM: @lonelycode See line 194. 

```
if maxBytes < collStats["size"] {
    // TODO Write warning log advising that resizing collection ignored due to data loss
    return
}
```

Proposed pseudocode logic is to ensure that `collStats["size"]` is smaller than requested maxBytes before resizing the capped collection to ensure no data loss. Maybe we should add a buffer (e.g. 10%) for safety.

Alternatively, as you suggest - we just don't resize capped collections and I remove this logic entirely.

PR #58, opened 2017-12-08
Add Splunk pump

matiasinsaurralde:
This is a simple wrapper that interacts with the Splunk HTTP Event Collector service, following the format that's specified [here](http://dev.splunk.com/view/event-collector/SP-CAAAE6P).

splunkpump

A sample configuration looks like this:
```json
    "pumps": {
        "splunk": {
            "name": "splunk",
            "meta": {
                "collector_token": "85FBC7DE-A1A1-A1A1-A1A1-A197D3510464",
                "collector_url": "https://input-prd-p-xyz.cloud.splunk.com:8088",
                "tls_skip_verify": true
            }
        }
    },
```

This is the result of some tweaks around my initial work on `go-splunk`, I've updated it a bit to take the timestamp from the actual request and attach it to the event in the way the HTTP Event Collector expects.
The TLS part needs some work, I'm thinking that the Splunk Cloud platform I'm using for tests provides a certificate that we can use.`tls_skip_verify` is useful for development purposes, for now.

### DIFF
--- a/pumps/init.go
+++ b/pumps/init.go
@@ -20,4 +20,5 @@ func init() {
 	AvailablePumps["statsd"] = &StatsdPump{}
 	AvailablePumps["segment"] = &SegmentPump{}
 	AvailablePumps["graylog"] = &GraylogPump{}
+	AvailablePumps["splunk"] = &SplunkPump{}
 }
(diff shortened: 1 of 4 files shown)


buger:
@matiasinsaurralde does it ready to be merged?

buger:
Also rename ssl verification option to `SSLInsecureSkipVerify` `ssl_insecure_skip_verify` to reflect our similar options in rest of products.

matiasinsaurralde:
@buger have added that, still finding out how Splunk manages the cert part, will let you know when it's finished.

PR #65, opened 2018-02-15
Update pump.conf to include mongo aggregates pump

joshblakeley:
This pump is required for users to see analytics graphs in their installs so should be present as default.

### DIFF
--- a/pump.example.conf
+++ b/pump.example.conf
@@ -1,34 +1,41 @@
 {
-	"analytics_storage_type": "redis",
+  "analytics_storage_type": "redis",
     "analytics_storage_config": {
-        "type": "redis",
-        "host": "localhost",
-        "port": 6379,
-        "hosts": null,
-        "username": "",
-        "password": "",
-        "database": 0,
-        "optimisation_max_idle": 100,
-        "optimisation_max_active": 0,
-        "enable_cluster": false
-    },
-    "purge_delay": 10,
-    "pumps": {
-        "mongo": {
-            "name": "mongo",
-            "meta": {
-                "collection_name": "tyk_analytics",
-                "mongo_url": "mongodb://localhost/tyk_analytics",
-                "collection_cap_max_size_bytes": 1048576,
-                "collection_cap_enable": true
-            }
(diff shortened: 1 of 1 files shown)


pump.example.conf
```diff
@@ -24,6 +24,13 @@
             }
         }
     },
+    "mongo-pump-aggregate": {
+    "name": "mongo-pump-aggregate",
+    "meta": {
+      "mongo_url": "mongodb://localhost/tyk_analytics"
+    },
+    "use_mixed_collection": true
```
excieve: This should be inside `meta` I think.

pump.example.conf
```diff
@@ -24,6 +24,13 @@
             }
         }
     },
+    "mongo-pump-aggregate": {
+    "name": "mongo-pump-aggregate",
```
excieve: Please fix the indentation.

pump.example.conf
```diff
@@ -24,6 +24,13 @@
             }
         }
     },
+    "mongo-pump-aggregate": {
```
excieve: This should be inside `pumps`.

pump.example.conf
```diff
@@ -24,6 +24,13 @@
...
+    	"name": "mongo-pump-aggregate",
+    	"meta": {
+      		"mongo_url": "mongodb://localhost/tyk_analytics"
+    	},
+    	"use_mixed_collection": true
```
excieve: This should be inside `meta`.

excieve:
Sorry for nitpicking but there are still indentation issues — probably a mix of spaces and tabs.

In addition to this, it seems the example config is using collection caps but the template does not. We should add them to the template too I guess.
@asoorm Is collection cap usage supposed to be in a typical pump installation as defined in the sample?

asoorm:
@excieve For sensible defaults - I put in example config file
@joshblakeley If we are fixing indentation in json config files... can we simply use json formatted with 2 spaces, rather than tabs?

Javascript is not really readable with tabbed indentation on github IMO.

PR #115, opened 2019-06-13
Kafka pump

Zavierazo:


Zavierazo:
@letzya fixed. Instead of use kafka with async mode i have declared every time the connection and close after use.

Zavierazo:
Hi @letzya @asoorm , we have finished our test on staging environment for this pump. Performance in general terms is much better than CsvPump + FileBeat + Kafka +LogStash

Currently with csvPump we have a cpu usage of 80-90% for tyk-pump.

With KafkaPump + LogStash with the same number of RPS(About 5k) the cpu usage is less than 10%.

We want to deploy this pump to our live env. How much time do you expect to have an rpm with these changes?

Regards

pumps/kafka.go
```diff
@@ -119,8 +124,8 @@ func (k *KafkaPump) WriteData(data []interface{}) error {
 	return nil
 }
 
-func (k *KafkaPump) write(messages []kafka.Message) error{
+func (k *KafkaPump) write(messages []kafka.Message) error {
 	kafkaWriter := kafka.NewWriter(k.writerConfig)
 	defer kafkaWriter.Close()
```
letzya: Are you sure we should close connection after every write call to kafka?

Zavierazo: In theory yes, with the defer operation when go out of write method kafka client is closed

Zavierazo: @letzya My previous version are without close conection and you ask me to close the connection. The only way i see currently with the pump is open and close connection on every batch. As i can see in other pump connection is opened for every batch. Then i assume we don't have any other way to define the close connection instead of open&close every batch.

letzya: I need to check that.

PR #136, opened 2019-08-09
Updating concurrent_map to latest from 3rd party to fix deadlock

mark-bradshaw:
At my company, we've been having a deadlock in the redis clustering code in the pump for some time.  Looking more deeply into it we determined the issue was in `vendor/github.com/TykTechnologies/redigocluster/rediscluster/concurrent_map.go`.  Considering the troubleshooting tip on the bottom of `https://tyk.io/docs/configure/redis-cluster/` it appears that this is a widespread issue.  

Doing a bit more digging I determined that the concurrent_map.go original project source (https://github.com/orcaman/concurrent-map/) has known about this issue and fixed it quite a while ago.  You can see the fix happening in this commit (https://github.com/orcaman/concurrent-map/commit/11703b488ea29b62cb1411eb41bcfc5976f1427a) conveniently labeled `Fix deadlock when using Iter/IterBuffered`.

Given that the concurrent map code in the tyk-pump appears to be years behind fixes in the upstream project I've pulled down their code and done some minor fixes to make it respect the shard count environment variable.  It's unclear if we even need to keep that variable if the deadlock issue is fixed, but I left it just in case there are other reasons to make increasing the shard count desirable other than dealing with the deadlock.

Please review this updated code and pull it in.  This is an issue that has broken the pump in production for my company.  Also, you might want to add proper attribution in that code since it's pulled from another project that has copyright.

### DIFF
--- a/vendor/github.com/TykTechnologies/redigocluster/rediscluster/concurrent_map.go
+++ b/vendor/github.com/TykTechnologies/redigocluster/rediscluster/concurrent_map.go
@@ -8,11 +8,10 @@ import (
 )
 
 const (
-	SHARD_COUNT     = 32
 	SHARD_COUNT_ENV = "REDIGOCLUSTER_SHARDCOUNT"
 )
 
-var customShardCount string
+var SHARD_COUNT = 32
 
 // A "thread" safe map of type string:Anything.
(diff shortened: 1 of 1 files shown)


mark-bradshaw:
I've moved this to https://github.com/TykTechnologies/redigocluster/pull/6

PR #140, opened 2019-08-14
Add option to obfuscate keys when writing to splunk

buger:
Splunk pump now has `obfuscate_api_keys` option

### DIFF
--- a/pumps/splunk.go
+++ b/pumps/splunk.go
@@ -144,10 +144,10 @@ func (p *SplunkPump) WriteData(data []interface{}) error {
 		apiKey := decoded.APIKey
 		if p.config.ObfuscateAPIKeys {
 			if len(apiKey) > 4 {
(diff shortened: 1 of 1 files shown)


pumps/splunk.go
```diff
@@ -140,11 +141,18 @@ func (p *SplunkPump) WriteData(data []interface{}) error {
 	}).Info("Writing ", len(data), " records")
 	for _, v := range data {
 		decoded := v.(analytics.AnalyticsRecord)
+		apiKey := decoded.APIKey
+		if p.config.ObfuscateAPIKeys {
+			if len(apiKey) > 4 {
+				apiKey = "****" + keyName[len(apiKey)-4:]
```
letzya, line 147: I believe keyName should be apiKey :)

tbuchaillot: updated

Issue #38, opened 2017-05-17
tyk_analytics Alias mapping


coatezy:
I'm pretty new to Tyk and Pump but after having a dig around what is stored in Redis and the mappings [here](https://github.com/TykTechnologies/tyk-pump/blob/master/analytics/analytics.go#L39) I was just wondering why `Alias` is not defined in the pump mappings?

I'd like to define our users ID in the alias so we can link analytics to our users.

Happy to open a PR if this is something you are open too, but thought I'd ask first as there my be a geniune reason why `Alias` has not been included.

yourpalal:
Hi, I wanted this feature as well, and added it to a fork, alnong with an option to exclude the api_key field from ES output. I didn't PR because of  the slightly hackish configuration method I used, but this change could easily be adapted for an alias-only change.

https://github.com/omxhealth/tyk-pump/commit/ca2c3babf398bf475eab14bb7b8a3174b3f47f4f

coatezy:
Thanks @yourpalal! That is exactly what I was planning on doing. Did you have any problems building? I am hitting the following errors.

```
go build -v ./...
vendor/github.com/garyburd/redigo/redis/pool.go:28:2: use of internal package not allowed
vendor/gopkg.in/mgo.v2/auth.go:38:2: use of internal package not allowed
vendor/gopkg.in/mgo.v2/bson/json.go:7:2: use of internal package not allowed
```

I've not got much experience with Go though so I may just be doing something wrong. 🙈

yourpalal:
Hmmm I don't have much experience with go either and it must have worked at the time, since I definitely built it. Trying again now gives me the same error as you, so I'm not sure what's up! sorry!

coatezy:
Good to know it's not just me. The last build on Travis also passed 28 days ago. It can't be a Go version issue as I've tried:

- 1.7.4_2
- 1.8
- 1.8.1

I'll report back if/when I get it building.

mvdan:
`tyk-pump` builds fine here on master on both 1.8.1 and tip. I think those kinds of issues arise when your Go installation is somehow messed up, but that probably isn't it if you're both hitting it.

Have you tried `go build .` or `go install .`, i.e. using `.` instead of `./...`?

mvdan:
Could also be because you guys are using the wrong import path for the build: https://stackoverflow.com/questions/34973883/how-do-i-fix-the-error-message-use-of-an-internal-package-not-allowed-when-go

[...]

PR #145, opened 2019-08-23
Fixed in Elasticsearch analytics to publish the alias

cvaram96:
This PR is for publishing the alias into the Elasticsearch. We can use the alias for different purposes such as identifying the users in the analytics.
 [Github issue](https://github.com/TykTechnologies/tyk-pump/issues/38) and [community forum ](https://community.tyk.io/t/user-identifier-in-analytic-logs/1621/3) are described this issue.


buger:
Looks good!

PR #147, opened 2019-08-27
Wait for network before starting

alephnull:
Only the systemd service


excieve:
Answering my own question: Prometheus pump launches an HTTP server, so that's fine.

asoorm:
Also, doesn't network need to be available in order to connect to Redis & Mongo?

excieve:
@asoorm Yep, but as long as it doesn't crash (keeps trying to re-connect) that's fine IMO. The less boot process is blocking the better usually.

PR #149, opened 2019-10-03
Fix pumping tcp analytics records

gernest:
This skips tcp analytics records. We still need to design
how tcp analytics records will look like

Fixes https://github.com/TykTechnologies/tyk/issues/2561


buger:
It is in master already

PR #160, opened 2019-10-17
[Requires Testing] non-blocking kafka write. info-level logging for time to write to Kafka

asoorm:
If there is a slowdown in Kafka:

1. this change should highlight the issue as info-logs as the time to write records will be recorded.
2. because it writes logs in separate go-routine, it will not block the next purge from redis meaning that we maintain stability of Tyk in the event of un-related issues.

### DIFF
--- a/pumps/kafka.go
+++ b/pumps/kafka.go
@@ -3,12 +3,13 @@ package pumps
 import (
 	"context"
 	"encoding/json"
+	"time"
+
 	"github.com/TykTechnologies/logrus"
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/mitchellh/mapstructure"
 	"github.com/segmentio/kafka-go"
 	"github.com/segmentio/kafka-go/snappy"
-	"time"
 )
 
 type KafkaPump struct {
@@ -74,6 +75,7 @@ func (k *KafkaPump) WriteData(data []interface{}) error {
 	startTime := time.Now()
 	k.log.Info("Writing ", len(data), " records...")
 	kafkaMessages := make([]kafka.Message, len(data))
+
 	for i, v := range data {
 		//Build message format
 		decoded := v.(analytics.AnalyticsRecord)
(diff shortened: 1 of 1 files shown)


pumps/kafka.go
```diff
@@ -104,23 +106,29 @@ func (k *KafkaPump) WriteData(data []interface{}) error {
...
-	//Send kafka message
-	kafkaError := k.write(kafkaMessages)
-	if kafkaError != nil {
-		k.log.WithError(kafkaError).Error("unable to write message")
-	}
-	k.log.Debug("ElapsedTime in seconds for ", len(data), " records:", time.Now().Sub(startTime))
+
+	// send kafka message in separate go-routine so that will not block the next Redis purge in event of Kafka slowdown.
+	// should only create new go-routine every `config.purge_delay` seconds
+	go func(startTime time.Time, log *logrus.Entry) {
```
letzya: looks good. Need testing on a heavy load of the pump

pumps/kafka.go
```diff
@@ -104,23 +106,29 @@ func (k *KafkaPump) WriteData(data []interface{}) error {
...
+		err := k.write(kafkaMessages)
+		if err != nil {
+			log.WithError(err).Error("unable to write message")
+		}
+
+		log.Infof("written %d records %s", len(data), time.Now().Sub(startTime).String())
```
letzya: fixed the message to refer to the time period 
"written %d records in %s seconds"

letzya: BTW, should we use _k.log.Time_ instead of _time.Now()_ ?

letzya: actually it's instead of passing the start time. since the log entry is by value

letzya: actually about _log *logrus.Entry_ - should it be a pointer? it's not in _func (entry Entry) log(level Level, msg string)_

asoorm: 1. the `String()` method automatically adds sensible units. So it will log: `written 100000 records 1.2s` or similar
2. I've never used log.Time - I think that referrs to the time the log was written
3. We shouldn't really be timing jobs here anyway, but actually using instrumentation.

```
job.Timing("purge_time_"+pmp.GetName(), time.Since(startTime).Nanoseconds())
```

asoorm:
```
// String returns a string representing the duration in the form "72h3m0.5s".
// Leading zero units are omitted. As a special case, durations less than one
// second format use a smaller unit (milli-, micro-, or nanoseconds) to ensure
// that the leading digit is non-zero. The zero duration formats as 0s.
```

letzya: About entry.Time: check the log function in entry.go

> func (entry Entry) log(level Level, msg string) {
> 	var buffer *bytes.Buffer
> 	entry.Time = time.Now()

PR #163, opened 2019-10-29
Update distro releases

vesko-tyk:
precise was EOL a couple of years ago
bionic is the newest LTS

### DIFF
--- a/utils/dist_push.sh
+++ b/utils/dist_push.sh
@@ -17,10 +17,12 @@ do
     rpmName="tyk-pump-$VERSION-1.${arch/amd64/x86_64}.rpm"
 
     echo "Pushing $debName to PackageCloud"
-    package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/precise $debName
+    package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/buster $debName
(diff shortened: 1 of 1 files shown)


utils/dist_push.sh
```diff
@@ -17,10 +17,12 @@ do
     rpmName="tyk-pump-$VERSION-1.${arch/amd64/x86_64}.rpm"
 
     echo "Pushing $debName to PackageCloud"
-    package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/precise $debName
+    package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/buster $debName
```
excieve: This should be `debian`, not `ubuntu`.

utils/dist_push.sh
```diff
@@ -17,10 +17,12 @@ do
...
     package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/trusty $debName
     package_cloud push tyk/$PACKAGECLOUDREPO/ubuntu/xenial $debName
     package_cloud push tyk/$PACKAGECLOUDREPO/debian/jessie $debName
+    package_cloud push tyk/$PACKAGECLOUDREPO/debian/stretch $debName
+    package_cloud push tyk/$PACKAGECLOUDREPO/debian/bionic $debName
```
excieve: And this should be `ubuntu` ;)

Issue #154, opened 2019-10-10
Gracefully truncate or ignore data when size is bigger than 16MB


ConsM:
At the moment this will result in an error
 `Problem inserting to mongo collection: BSONObj size: X (0x28CB7F0) is invalid. Size must be between 0 and 16793600(16MB)`

dencoded:
It seems like truncating data (as suggested in the title of this issue might be too agressive)

According to our implementation config param `max_insert_batch_size_bytes` should prevent sending data sets more than specified (default value `10MB`):
https://github.com/TykTechnologies/tyk-pump/blob/master/pumps/mongo.go#L344

Looks like it is not calculated correctly and tries to sends more which causes this issue.

I think we have a bug here: https://github.com/TykTechnologies/tyk-pump/blob/master/pumps/mongo.go#L330
we add size of raw request to itself, however response can be much bigger.

Looks like it was supposed to be size of raw request + size of raw response:
`len([]byte(thisItem.RawRequest)) + len([]byte(thisItem.RawResponse))`

asoorm:
What @dencoded says makes sense. Also no need to do byte slice conversion, given that strings are immutable byte slices.

Separate issue - we should probably store `RawRequest` and `RawResponse` as byte slices to avoid copying & reduce memory allocations.

letzya:
Btw, A client  got the same error in the aggregated pump as well.

furkansenharputlu:
@asoorm @matiasinsaurralde Can you please see my solution?

PR #165, opened 2019-10-30
Calculate document size correctly and don't skip to write some useful data

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk-pump/issues/154

### DIFF
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -2,18 +2,17 @@ package pumps
 
 import (
 	"crypto/tls"
-	"net"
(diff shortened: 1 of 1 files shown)


asoorm:
We could export a new field? `raw_output_omitted: bool`. We also store content_length - so it should be obvious if that number is particularly large?

buger:
@asoorm makes sense, I'll merge it like this for now, and for 2.10 we plan to move out raw data to separate collection, so we will address your comment there.

PR #166, opened 2019-11-04
Add config option to ignore tags with common prefix

### LINKED ISSUE #162: Provide an option to ignore specific tags from aggregated analytics
Mongo aggregate pump should provide a way to ignore specific tags from analytics. 
This is useful when `tag_headers` are used for most of APIs, which might result in exceed of mongo document size limit.

komalsukhani:
Fixes #162 

Added additional config options:
1. `ignore_tag_prefix_list`( in mongo aggregate and hybrid pump.)
It will not store analytics for tags having prefix specified in the list.
**Note**: Prefix `key-` is added in the list by default. This tag is added by gateway for keys.

2. `threshold_len_tag_list`( in mongo aggregate pump)
If number of tags in a document grows beyond `threshold_len_tag_list`, pump will throw a warning. The warning will print top 5 common tag prefix. Default value is **1000**. To disable alerts set it to -1.

### DIFF
--- a/analytics/aggregate.go
+++ b/analytics/aggregate.go
@@ -353,8 +353,18 @@ func doHash(in string) string {
 	return search
 }
 
+func ignoreTag(tag string, ignoreTagPrefixList []string) bool {
+	for _, prefix := range ignoreTagPrefixList {
+		if strings.HasPrefix(tag, prefix) {
+			return true
+		}
+	}
+
+	return false
+}
+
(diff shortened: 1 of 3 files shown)


sedkis:
@buger @komalsukhani 
Can we please document these by adding mongo aggregate example to tyk pump readme?

buger:
Makes sense...

PR #167, opened 2019-11-04
Option to store analytics per minute

### LINKED ISSUE #164: Aggregate analytics document size grows more than 16MB
Currently, aggregate data is generated on hourly basis. That could be one the reason.
Aggregate pump should provide an option to record data per minute.

komalsukhani:
Fixes #164 
Added `store_analytics_per_minute` config option in aggregate and hybrid pumps.

Currently, we generate aggregate data per hour. If this option is enabled, aggregate data will be generated per minute.

### DIFF
--- a/pumps/hybrid.go
+++ b/pumps/hybrid.go
@@ -38,8 +38,9 @@ var (
 
 // HybridPump allows to send analytics to MDCB over RPC
 type HybridPump struct {
-	aggregated    bool
-	trackAllPaths bool
+	aggregated             bool
+	trackAllPaths          bool
+	storeAnalyticPerMinute bool
 }
 
 func (p *HybridPump) GetName() string {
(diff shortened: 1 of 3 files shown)


pumps/hybrid.go
```diff
@@ -38,8 +38,9 @@ var (
 
 // HybridPump allows to send analytics to MDCB over RPC
 type HybridPump struct {
-	aggregated    bool
-	trackAllPaths bool
+	aggregated             bool
+	trackAllPaths          bool
+	storeAnalyticPerMinute bool
```
buger: What this value is for?

komalsukhani: If this is enabled, analytics will be stored per minute. Otherwise, analytics will be stored per hour, which is current behaviour

PR #171, opened 2019-11-05
Add prefix to logs for production deployments

sredxny:
This branch solves https://github.com/TykTechnologies/product/issues/236  was updated as well the Tyk's repos related to logs, and this should be merged in parallel with the next Pull requests as this depends and use the code contained there:
- https://github.com/TykTechnologies/logrus-prefixed-formatter/pull/1
- https://github.com/TykTechnologies/logrus/pull/1

**What was done?**
Forced the logger to use Colors so, the information related to the prefix is not lost when they are printed in logs files. Was written a test in _logger/init_test.go_ to check that the prefix is printed in the log file.

**How was tested locally?**
- Ran tyk-pump with: `go build -i && ./tyk-pump --conf ../my-pump.conf 2>&1 | logger -s -t $(basename $0)4`
- From another terminal, check the latest logs: `journalctl -f`
- Checked that we have now the prefix printed in logs with no TTY: 
Captura de pantalla de 2019-11-05 12-10-41

**About the test**
Consist in writting the logs in a file, after that we check if the logs contains the word "prefix" is so, then it Pass, otherwise its a Failed test


buger:
@sredxny pls fix the conflicts and can be merged.

PR #178, opened 2019-11-19
Handle unsupported MongoDB characters

### LINKED ISSUE #113: If endpoint path contains dot, then mongodb data gets corrupted
Analytics data for an endpoint is recorded in a field with name same as that of the path.
MongoDB doesn't support dot in the field name. If the path contains a dot, mongodb collection gets corrupted.

**Example**

If analytics is enabled for endpoint `/get/(.*)`, instead of creating field `/get/(.*)`, `/get/( ` embedded document with `*)` field gets created.

komalsukhani:
Fixes #113 

MongoDB doesn't support `.` in parent field names.[[1](https://docs.mongodb.com/v3.4/reference/limits/#Restrictions-on-Field-Names)]
If an endpoint path contained `.` , mongoDB data was corrupted.
To fix this, `.` char will be replaced with it's unicode representation.

### DIFF
--- a/analytics/aggregate.go
+++ b/analytics/aggregate.go
@@ -2,6 +2,7 @@ package analytics
 
 import (
 	b64 "encoding/base64"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


buger:
👍

Issue #177, opened 2019-11-18
Enable "log_level" in "pump.conf"


matiasinsaurralde:
Similar to https://github.com/TykTechnologies/tyk/issues/1182, seems it's not currently possible to do this with Pump

PR #179, opened 2019-11-22
Extend logging settings

matiasinsaurralde:
For #177 

This introduces the following enhancements:
* `--debug` is flag is added.
* It's possible to modify log level using `pump.conf`, using the `log_level` flag, just like the gateway.
* In case the log level is modified it affects all pumps (they share the same logger object).


matiasinsaurralde:
@sredxny this could be a good enhancement after #171 is merged 🙂

sredxny:
@matiasinsaurralde Good idea. I checked the changes and so far they seems to be very well to me

PR #180, opened 2019-11-23
Add support for multiple ES urls

buger:
The underline client already supports having multiple URLs, our job is just to provide it.
At the moment implementation rely on "," separator.


buger:
🙈

PR #181, opened 2019-11-24
Extend ES pump with `decode_base64` boolean option

buger:
Allows decoding base64 raw request/response data.


christtyk:
@buger can we add to our next pump release if space pls

joshblakeley:
Which release will this be out with?

buger:
@joshblakeley it was part of last pump release 0.8.4

Issue #117, opened 2019-06-18
Uptime default mongo db connection should provide better logs or exit


theholy7:
https://github.com/TykTechnologies/tyk-pump/blob/0b9147421f79b5542390c7b9d47f4fabe5f7135b/pumps/mongo.go#L264

I was using a config to send logs to elasticsearch, the only pump I had setup, and without the `dont_purge_uptime_data` flag set.

By default it creates a MongoDB pump, for which I didn't have a MongoDB server setup.

tyk-pump was hanging in the mongo Pump connect, and even with info/debug log levels, the error message never appeared, so I couldn't track down the error.

I couldn't find this properly documented, and the logs never showed up for me when I was running the docker container. Below is a copy paste.

```
[Jun 18 15:45:15]  INFO main: ## Tyk Analytics Pump, v0.6.0 ##
[Jun 18 15:45:15]  INFO elasticsearch-pump: Elasticsearch URL: https://user:pass@elasticsearch:esport * removed secrets
[Jun 18 15:45:15]  INFO elasticsearch-pump: Elasticsearch Index: tyk_analytics
[Jun 18 15:45:15]  INFO main: Init Pump: Elasticsearch Pump
[Jun 18 15:45:15]  INFO mongo-pump: -- No max batch size set, defaulting to 10MB
[Jun 18 15:45:15]  INFO mongo-pump: -- No max document size set, defaulting to 10MB
```

PR #188, opened 2019-12-13
improve uptime pump by adding more logs and removing the infinite loop

sedkis:
currently the timeout is infinite, and there is no logs.  this is a poor user experience
this improves uptime pump by adding more logs and removing the infinite loop.

fixes https://github.com/TykTechnologies/tyk-pump/issues/117

### DIFF
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -261,10 +261,12 @@ func (m *MongoPump) connect() {
 	dialInfo.Timeout = time.Second * 5
 	m.dbSession, err = mgo.DialWithInfo(dialInfo)
 
-	if err != nil {
+	for err != nil {
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
-		}).Fatal("Mongo connection failed:", err)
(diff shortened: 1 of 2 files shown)


pumps/mongo.go
```diff
@@ -258,16 +258,13 @@ func (m *MongoPump) connect() {
...
 	m.dbSession, err = mgo.DialWithInfo(dialInfo)
 
-	// TODO - Should this not bail after a while?
-	for err != nil {
+	if err != nil {
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
-		}).Error("Mongo connection failed:", err)
-
-		time.Sleep(5 * time.Second)
```
buger: Previously it could auto-heal by itself since it was inside the loop, now it will fail on first run. Is that expected? Not sure we should change this behavior.

sedkis: well, I can change it to print error message and retry connection instead of failing?

sedkis: I fixed it!

Issue #189, opened 2019-12-16
Adding a Pump for Syslog


sedkis:
I want my pump to spit out my request and response logs to the standard output, or syslog.

I want a single entry for each request.   is this possible?  what would be the throughput capabilities?

adelowo:
If a pump for standard output is going to be implemented, let's make it fluentd compatible.

sedkis:
keep open!

letzya:
@sedkis what's the benefit of it over csv? is it for dev env in which you want to quickly see the records or maybe something else? worse adding the use case and the necessity so it can be prioritise :)

sedkis:
pr in progress

PR #191, opened 2019-12-20
[TN-5] Add syslog pump.

sedkis:
For #189 

Docs: https://github.com/TykTechnologies/tyk-docs/pull/1005

Fix [TN-5]

[TN-5]: https://leotyktest.atlassian.net/browse/TN-5


pumps/syslog.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"github.com/TykTechnologies/logrus"
+	"github.com/TykTechnologies/tyk-pump/analytics"
+	"github.com/mitchellh/mapstructure"
+	"log/syslog"
+)
+
+type SyslogPump struct {
+	syslogConf *SyslogConf
+	Writer     *syslog.Writer
```
adelowo: It will be better not to export this.

adelowo:
```suggestion
      writer     *syslog.Writer
```

pumps/syslog.go
```diff
@@ -0,0 +1,140 @@
+package pumps
+
+import (
+	"fmt"
+	"github.com/TykTechnologies/logrus"
+	"github.com/TykTechnologies/tyk-pump/analytics"
+	"github.com/mitchellh/mapstructure"
+	"log/syslog"
+)
```
adelowo: can this be properly formatted?

pumps/syslog.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s *SyslogPump) Init(config interface{}) error {
+	//Read configuration file
+	s.syslogConf = &SyslogConf{}
+	err := mapstructure.Decode(config, &s.syslogConf)
+
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"prefix": logPrefix,
+		}).Fatal("Failed to decode configuration: ", err)
```
matiasinsaurralde: Should we `return` here?

adelowo: We don't need to. `Fatal` will shut down the process

sedkis: I prefer to shut down the pump if the desired pumps aren't loading correctly

pumps/syslog.go
sedkis: sorry how do you mean @adelowo ?

sedkis: ahh you mean with spaces! 
done now

sedkis:
@buger , 

is this PR missing anything?  I can update the pump README and then I'll create a PR for the docs

buger:
If our final goal is fluentD, why not to make it compatible?

sedkis:
> If our final goal is fluentD, why not to make it compatible?

Here is where I'm struggling because Syslog is going to have a different implementation depending on the OS.  I'm fairly certain that FluentD users will regularly need to provide a parsing algorithm to properly insert Logs

sedkis:
Added syslog sections to the Pump readme as well as the Tyk Docs

pumps/syslog.go
```diff
@@ -0,0 +1,142 @@
...
+	log.Debug("Syslog Pump active")
+	return nil
+}
+
+func initWriter(s *SyslogPump) {
+	syslogWriter, err := syslog.Dial(
+		s.syslogConf.Transport,
+		s.syslogConf.NetworkAddr,
+		syslog.Priority(s.syslogConf.LogLevel),
+		"syslog-pump")
```
letzya: tag could be configurable and defaulting to `logPrefix` 's value

tbuchaillot: Added `tag` to config options.

PR #195, opened 2020-01-06
Add SSL option for Kafka connection

### LINKED ISSUE #194: SSL option for Pump-Kafka connection
Right now, it is not possible to make connection between Pump and Kafka secure with SSL. The option should be implemented.

furkansenharputlu:
Added new parameters to Kafka config:
- `use_ssl`
- `ssl_insecure_skip_verify`

Fixes #194

### DIFF
--- a/pumps/kafka.go
+++ b/pumps/kafka.go
@@ -2,6 +2,7 @@ package pumps
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"github.com/TykTechnologies/logrus"
 	"github.com/TykTechnologies/tyk-pump/analytics"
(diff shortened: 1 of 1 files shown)


letzya:
@furkansenharputlu can we add #235 to this PR?
 Something like 
```	var tlsConfig *tls.Config
	var cert tls.Certificate
	if k.kafkaConf.UseSSL {
		if k.kafkaConf.SSLCertFile != "" && k.kafkaConf.SSLKeyFile != "" {
			// write a debug message that loading  certificates if mtls is set in kafka 
			cert, err = tls.LoadX509KeyPair(k.kafkaConf.SSLCertFile, k.kafkaConf.SSLKeyFile)
			if err != nil {
				return err
			}
		} else {
                      if (k.kafkaConf.SSLCertFile != "" || k.kafkaConf.SSLKeyFile != "" ) {
                        //write a message that only one of configs was set so mtls won't work
                     }
                }
	} 
       
	tlsConfig = &tls.Config{
		Certificates: []tls.Certificate{cert},
		InsecureSkipVerify: k.kafkaConf.SSLInsecureSkipVerify,
	}
```

PR #200, opened 2020-01-09
Added message to clarify purging state.

letzya:
With a nonce in case the flush to the log is out of sync or if we change this to go routines

### DIFF
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -297,9 +297,10 @@ func (m *MongoPump) WriteData(data []interface{}) error {
 
 			analyticsCollection := sess.DB("").C(collectionName)
 
+			nonce := time.Now() //needed in case the flush to the log is out of sync
 			log.WithFields(logrus.Fields{
 				"prefix": mongoPrefix,
-			}).Info("Purging ", len(dataSet), " records")
+			}).Info("Purging ", len(dataSet), " chunks (for nonce: ", nonce, ").")
 
 			err := analyticsCollection.Insert(dataSet...)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pumps/mongo.go
```diff
@@ -297,17 +297,22 @@ func (m *MongoPump) WriteData(data []interface{}) error {
 
 			analyticsCollection := sess.DB("").C(collectionName)
 
+			nonce := time.Now() //needed in case the flush to the log is out of sync
 			log.WithFields(logrus.Fields{
 				"prefix": mongoPrefix,
-			}).Info("Purging ", len(dataSet), " records")
+			}).Info("Purging ", len(dataSet), " chunks (for nonce: ", nonce, ").")
```
furkansenharputlu: I think this is wrong. They are `len(dataSet)` records but just `1` chunk. So it should stay same. Also, it will log the the time, why do we add time again ?

pumps/mongo.go
```diff
@@ -297,17 +297,22 @@ func (m *MongoPump) WriteData(data []interface{}) error {
...
 			if err != nil {
 				log.Error("Problem inserting to mongo collection: ", err)
 				if strings.Contains(strings.ToLower(err.Error()), "closed explicitly") {
 					log.Warning("--> Detected connection failure!")
 				}
+				return
 			}
+			log.WithFields(logrus.Fields{
+				"prefix": mongoPrefix,
+			}).Info("Have purged ", len(dataSet), " chunks (for nonce: ", nonce, ").")
```
furkansenharputlu: same here.

pumps/mongo.go
```diff
@@ -297,17 +297,22 @@ func (m *MongoPump) WriteData(data []interface{}) error {
 
 			analyticsCollection := sess.DB("").C(collectionName)
 
+			nonce := time.Now() //needed in case the flush to the log is out of sync
```
furkansenharputlu: It will log the the time, why do we add time again ?

buger:
I believe this idea was not fully implemented yet

PR #201, opened 2020-01-10
Create SSLConf struct to use in Kafka

furkansenharputlu:
This PR creates an `SSLConfig` struct because it is scalable.

### DIFF
--- a/pumps/kafka.go
+++ b/pumps/kafka.go
@@ -23,14 +23,18 @@ type Json map[string]interface{}
 var kafkaPrefix = "kafka-pump"
 
 type KafkaConf struct {
-	Broker                []string          `mapstructure:"broker"`
-	ClientId              string            `mapstructure:"client_id"`
-	Topic                 string            `mapstructure:"topic"`
-	Timeout               time.Duration     `mapstructure:"timeout"`
-	Compressed            bool              `mapstructure:"compressed"`
-	MetaData              map[string]string `mapstructure:"meta_data"`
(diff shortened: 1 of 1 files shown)


furkansenharputlu:
Not stale

letzya:
Doesn't it breaks backward compatibility?

furkansenharputlu:
This is implemented with Kafka ssl feature but it is forgotten to be merged I think

pumps/kafka.go
```diff
@@ -54,9 +58,9 @@ func (k *KafkaPump) Init(config interface{}) error {
 	}
 
 	var tlsConfig *tls.Config
-	if k.kafkaConf.UseSSL {
+	if k.kafkaConf.SSLConf.Enabled {
```
letzya, line 61: The pump is not a server. Why do we need to wrap InsecureSkipVerify inside this flag? Maybe we should use this flag if we also load certificates so the `SSLConf.Enabled` role is to turn on/off the mtls, but even for that, I'm not sure it's that necessary.

furkansenharputlu:
`InsecureSkipVerify` should be set on client side. Pump starts a `Kafka` client. And the client decides whether it should verify server's certificate or not.

tbuchaillot:
@furkansenharputlu can we close this PR? IMO it's going to break backward compatibility.

furkansenharputlu:
Yes, it is late to merge this PR from now on

Issue #204, opened 2020-01-22
Support ElasticSearch batch operations


tbuchaillot:
Optimize the way we write in ES. 

The idea is changing the way we write now (for loop + sending one by one record) for the ES bulk API to send a batch of analytics records in one roundtrip.

tbuchaillot:
Related #205

maciejwojciechowski:
can we make pump logs more clear? currently I have ES `"bulk_actions":10` but with every request I send I see in logs
```
INFO elasticsearch-pump: Writing 1 records
```
I think with such settings it would make sense to log INFO every 10 records.
Is it easy to improve pump logging for this feature?

maciejwojciechowski:
log problem will be handled separately

maciejwojciechowski:
verified

marksou:
When documenting the Batch features, I've spotted that we have miss-spelt `workers`. We have named it `wokers`.

PR #203, opened 2020-01-20
TT-613 limit number of records which pump pulls from redis list GetAndDeleteSet

asoorm:
Pending full QA

Further addresses: https://github.com/TykTechnologies/tyk-pump/issues/185

Traditionally, we pull the entire analytics list from Redis and delete the set. This is problematic for several reasons:

* If pump fails, **all** analytics are lost after TTL on analytics key expires.
* If multiple pumps are awaiting analytics, some pumps will get no records - so load is unevenly distributed
* If data-sink is slow, then analytics data elements build up in Redis. The pump will then pull ALL elements from the Redis List, and then try to ship all analytics - causing further pressure on the data-sink. Imagine sending 500k analytics records at 1kb each to ElasticSearch for ingestion for example.

---

This PR addresses the problems above by modifying the means by which we pull analytics records from Redis.

* new config option specifying the maximum number of records to pull from Redis at a time
* If PurgeChunk is 0 or unset, we will continue with current operation of pulling all analytics from Redis.
* If PurgeChunk is 100, this means we will pull 100 records at a time from Redis, meaning that we have more control over the chunk-size that we ship to upstream data-sources in one go.
* If PurgeChunk > 0, (enabled), then we check StorageExpirationTime. If not set, default to 60s. StorageExpirationTime will reset the TTL on the analytics record to StorageExpirationTime after a chunk has been purged. This means, that if the pump is not able to purge any records within 60s, then the analytics inside redis will expire. Effectively moving responsibility for resetting TTL on analytics record from the Gateway to the Pump - the entity which is purging the records.

## Considerations:

1. If data-sink is particularly slow, then there is a chance that the TTL inside Redis times out, and All analytics for that period will expire. This is expected functionality - because we don't want too many records building up in Redis - causing pressure on the DB.

2. Rather than TTL functionality inside Redis - maybe it would be more sensible / offer the option of controlling data-set size in Redis by using a kind of capped collection rather than a TTL?

3. Used in combination with PurgeDelay, We can now control & have visibility of how many RPS a single pump will purge.

```
PurgeDelay: 5
PurgeChunk: 10000
```

Means that the pump is configured to ingest & ship 2000 rps worth of analytics. And the data-sink(s) needs to be able to injest batches of 10k records. It becomes significantly easier to reason about & provides more guidance on configuration & sizing for pump deployments.

```
--- PASS: TestRedisClusterStorageManager_GetAndDeleteSet (0.08s)
[...]

PR #205, opened 2020-01-23
ES BulkProcessor

tbuchaillot:
Adding a new way of writing on ES,  ES BulkProcessor.

New config inside ES Meta:
```go
       "disable_bulk": false, 
        "bulk_config":{
            "wokers": int,             //number of wokers
            "bulk_actions":int ,   //number of requests
            "flush_interval":int,  //time in seconds
            "bulk_size": int        //ammount of bytes
        }
```

When any of these conditions are satisfied,  the batch of analytics records is sent to the ES.

Solves issue #204


tbuchaillot:
WIP Readme

buger:
Can we have some good defaults?

tbuchaillot:
@buger I think the best idea is making it work like the ES Bulk API when none of the bulk_conf are set. 
Maybe when #203 is merged we can use the `chunkSize` like the default `bulk_actions` number?

pumps/elasticsearch.go
```diff
@@ -62,14 +73,99 @@ func getOperator(version string, url string, setSniff bool) (ElasticsearchOperat
...
+
 		return e, err
 	case "6":
 		e := new(Elasticsearch6Operator)
 		e.esClient, err = elasticv6.NewClient(elasticv6.SetURL(urls...), elasticv6.SetSniff(setSniff))
+		if err != nil {
+			return e, err
+		}
+		// Setup a bulk processor
+		p := e.esClient.BulkProcessor().Name("TykPumpESv6BackgroundProcessor")
```
letzya: Any chance we can avoid triplicating the code? 33

Issue #92, opened 2019-02-25
All pumps should be configurable with env vars


excieve:
In some setups (e.g. running containers) it might be easier to do configuration through environment variables. We already support that for MongoDB pumps using `PMP_MONGO`, `PMP_MONGOAGG`, etc. env prefixes but that's missing from the ElasticSearch pump and possibly some others. This needs to be reviewed and implemented for all the pumps.

buger:
I guess it needs to be not only configurable, but also enabled by env vars as well. So if default config does not include ES, we set env vars for ES and it gets turned on.

I believe we can do it automatically for all pumps at once. We know the pre-defined name of each pump, so prefix will be always: `PMP_{upper(name)}`, so we need to check if env variables with such prefix exist, based on that "enable"/"auto-create" related pumps, and call envconfig before initializing each pump.

eloo:
Hello, 
is the usage of PMP_MONGO and others somewhere documented? Because i've not seen this in the documentation already.
thx

theholy7:
@excieve , what is the status of this issue? How do you recommend configuring an elastic search pump if it has basic auth, and you don't want to commit it to a repo?

asoorm:
It is currently possible to have multiple pumps of the same type. How could we support environment variables, whilst still allowing configuration of multiple pumps of same type?

e.g. `PMP_CSV_1_CSVDIR`, `PMP_CSV_2_CSVDIR`, `PMP_CSV_3_CSVDIR`? or would we simply only allow single pump type if Env vars are used?

Related: https://github.com/TykTechnologies/tyk-pump/issues/148

jlucktay:
From the [`envconfig` README](https://github.com/kelseyhightower/envconfig#supported-struct-field-types), it seems to me that since the pump configs are stored in a map like `map[string]PumpConfig` if the `PumpConfig` struct were to implement the `encoding.TextUnmarshaler` interface (which is also on `envconfig`s supported types list) then that would cover some of this issue off.

eloo:
i guess this is still open

jlucktay:
The [custom decoders](https://github.com/kelseyhightower/envconfig#custom-decoders) in `envconfig` would be another option to implement these with.

drpebcak:
Would love to see this supported because without it, tyk-pump is hard/impossible to configure properly when using the helm chart.

asoorm:
To give a concrete use-case for this in a k8s setup:

[...]

PR #230, opened 2020-03-04
Add dial info to Mongo connection error logs.

jlucktay:
Helps with feedback regarding issue #92.

- it is possible to specify settings for Mongo connections via
environment variables that are prefixed appropriately
- these log changes will show the dial info used by the Mongo connection
attempt so that any pertinent config changes can be seen, to aid
troubleshooting
- if the 'password' or other similarly sensitive fields on the dial info
struct were populated, they would show up in the error logs

For example, launching the pump with `PMP_MONGO_MONGOURL` set to `unavailable_url` would result in Mongo reporting the connection failure and (as a result of this change) including the bad URL in the error log, like so:
```
[Mar  4 10:09:57] ERROR mongo-pump: Mongo connection failed. Retrying. dialinfo=&{Addrs:[unavailable_url] ...
```

### DIFF
--- a/pumps/mongo.go
+++ b/pumps/mongo.go
@@ -418,7 +418,7 @@ func (m *MongoPump) connect() {
 	for err != nil {
 		log.WithFields(logrus.Fields{
 			"prefix": mongoPrefix,
-		}).Error("Mongo connection failed. Retrying. Err::", err)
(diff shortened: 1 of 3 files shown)


buger:
Makes sense

letzya:
perfect!

letzya:
since we dont have a test I like the fact that you added the log line as a proof.

PR #242, opened 2020-04-06
Implement debian origin signing

alephnull:
This was missed.


alephnull:
Please also merge into release branches.

PR #255, opened 2020-05-21
Feature/prometheus: Enrich Prometheus scraping endpoint

asoorm:
General Prometheus Metrics enhancements

- used base units (seconds) rather than (milliseconds)
- introduced histograms for `gateway latency`, `total latency`, `upstream latency`
- more granular metrics sharding, by api name & api version
- histogram buckets are now configurable
- metrics to expose are now configurable

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,15 +2,10 @@ package main
 
 import (
 	"context"
+	"os"
 	"strings"
 	"sync"
 	"time"
-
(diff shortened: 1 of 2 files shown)


tbuchaillot:
@asoorm We are looking to include full dynamic Prometheus metrics #431. Do you think we should close this?

PR #260, opened 2020-05-27
[TN-8] Healthcheck endpoint issues

### LINKED ISSUE #256: bug: healthcheck endpoint issues
https://tyk.io/docs/tyk-configuration-reference/tyk-pump-configuration/#health-check

1. Default port is same port as the gateway - that makes no sense - as this means the pump will not be able to bind to 8080 if the gateway is running, or the gateway won't start if the pump is running (and both are on same machine).

2. Listens on Localhost - meaning that K8s or in docker - you can't call the health endpoint.

3. We spin up a go routine - and can't cancel / stop it.

```
go server.ServeHealthCheck(SystemConfig.HealthCheckEndpointName, 
SystemConfig.HealthCheckEndpointPort)
```

4. We don't catch the error from http handler or do anything with it - so if the pump healthcheck tries to start on port 8080, cannot bind because that port is owned by gateway, then we don't know about it. We think the pump is down. 

https://github.com/TykTechnologies/tyk-pump/blob/53720da433bc22d881292d4eb6b778bd698ef736/server/server.go#L35

tbuchaillot:
## Description

Changing default healthcheck endpoint port to 8083 so it doesn't have a conflict with the gateway.
Removing localhost from ListenAndServe so it can work on Docker / k8s. If host is omitted it listen on all available interfaces instead.
Adding an error check on ListenAndServe so if the port is taken, the pump fails to initialize.

I updated the README but it needs to be updated in the documentation: https://github.com/TykTechnologies/tyk-docs/blob/0d05d16c09be9231b0bfa38cabc310b8d4eee6c1/tyk-docs/content/tyk-configuration-reference/tyk-pump-configuration/tyk-pump-configuration.md

## Related Issue

Closes #256 

## How This Has Been Tested

Locally - Docker.

## Types of changes

## Checklist

  - [x] If you've changed APIs, describe what needs to be updated in the documentation.
  - [x] `go fmt -s`
  - [x] `go vet`

## Testing instructions

- Setup a basic Tyk Pump installation.
- Start Tyk Pump with the default configuration values.
- Send a request to `http://localhost: 8080/hello` (default health check endpoint, with default port).
- Modify `tyk_pump.conf` with the following setting: `"health_check_endpoint_port": 8083`.
- Start Tyk Pump again and send a request to `http://localhost: 8083/hello`, expecting HTTP 200.


marksou:
I've changed the Docs to reflect this.

christtyk:
@tbuchaillot will this be included in next release?

tbuchaillot:
@christtyk yes

PR #267, opened 2020-06-23
Fix health endpoint to be published outside server.

### LINKED ISSUE #266: Health Check endpoint not accessible outside server
Currently if we made a call to http://localhost:8080/health works fine but if we made the same request from outside of the server doesn't work http://XXX.XXX.XXX.XXX:8080/health

Zavierazo:
## Description
Fix healthCheck to be accessible from outside of the server. Currently is only accessible from localhost

## Related Issue
fixes #266

## Motivation and Context
Be able to check health of application from outside with monitoring tools.

## How This Has Been Tested
Locally test with
http://localhost:8080/health
http://10.160.255.103:8080/health


buger:
Totally makes sense!

PR #270, opened 2020-07-01
Add new filters based on response_code's

Zavierazo:
Add new filters based on response_code's

## Description
Currently we have the necessity to be able to only send with pump request with errors. For that we need to implement this new filter to be able to remove all 200 response code.

## Motivation and Context
Decrease space required to store all data of analytics . Request with 200 are not important to be monitorized and also are the 99% of traffic. We have some problems regarding sizing of our cluster of kafka and we want to reduce the load.

## How This Has Been Tested
-Unit test
-Start pump in local with same config of our staging environment with the new filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ This feature adds a new configuration field in each pump called filters and its 
   "skip_response_codes":[]
 }
 ```
-The fields api_ids, org_ids and response_codes works as whitelists (APIs and orgs where we want to send the analytics records) and the fields skip_api_ids, skip_org_ids and skip_response_codes works as the opposite (blackslits).
+The fields api_ids, org_ids and response_codes works as allow list (APIs and orgs where we want to send the analytics records) and the fields skip_api_ids, skip_org_ids and skip_response_codes works as block list.
 
 The priority is always blacklisted configurations over whitelisted.
 
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ This feature adds a new configuration field in each pump called filters and its 
 ```
 The fields api_ids, org_ids and response_codes works as allow list (APIs and orgs where we want to send the analytics records) and the fields skip_api_ids, skip_org_ids and skip_response_codes works as block list.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -238,6 +238,37 @@ Create a `pump.conf` file:
...
+"filters":{
+  "api_ids":[],
+  "org_ids":[],
+  "response_codes":[],
+  "skip_api_ids":[],
+  "skip_org_ids":[],
+  "skip_response_codes":[]
+}
+```
+The fields api_ids, org_ids and response_codes works as whitelists (APIs and orgs where we want to send the analytics records) and the fields skip_api_ids, skip_org_ids and skip_response_codes works as the opposite (blackslits).
```
letzya, line 254: Could you please change the terms `whitelists` to allow list and and `blacklist` to block list?

letzya:
LGTM except for the terminology which we decided to update in Tyk but haven't posted it in the guideline of the PRs

README.md
Zavierazo: Done. Remember to update your documentation https://tyk.io/docs/tyk-configuration-reference/tyk-pump-configuration/tyk-pump-configuration/#configuration-1 I have based my description on that

letzya: thanks I think there's a PR for that :) cc @marksou

README.md
```diff
@@ -238,6 +238,37 @@ Create a `pump.conf` file:
...
+  "org_ids":[],
+  "response_codes":[],
+  "skip_api_ids":[],
+  "skip_org_ids":[],
+  "skip_response_codes":[]
+}
+```
+The fields api_ids, org_ids and response_codes works as allow list (APIs and orgs where we want to send the analytics records) and the fields skip_api_ids, skip_org_ids and skip_response_codes works as block list.
+
+The priority is always blacklisted configurations over whitelisted.
```
letzya, line 256: @ Zavierazo here as well

Zavierazo: Done

PR #271, opened 2020-07-01
Add capability to obfuscate keys for all records

letzya:
## Description

While filtering the record I thought it's good timing to obfuscate the key. For every record, I check if it needs to be filtered out and then obfuscate the key.
There's a global config field to turn this on - `obfuscate_keys`
Added TestObfuscateKeys just for that obfuscation function.
Added tests for filter+obfuscation of keys in a record
**Important: This PR is only useful if keys are not hashed, otherwise there's no real reason to obfuscate the key.**

Note: Removing the key from raw request is a bit tricky (need to get the name of the header to obfuscate and delete between the header to the end of the line- could be heavy operation) and best to solve this problem in the gateway and not record it in the first place

## Motivation and Context

Sometimes clients don't want to the API key to be recorded.
Important to note that it's not enough 

## How This Has Been Tested

Wrote unit tests

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


main.go
```diff
@@ -231,7 +228,11 @@ func filterData(pump pumps.Pump, keys []interface{}) []interface{} {
 		if filters.ShouldFilter(decoded) {
 			continue
 		}
-		filteredKeys[newLenght] = key
+		if !SystemConfig.RecordKeys {
```
tbuchaillot, line 231: @letzya, is it ok to change the default behavior? What if instead of using RecordKeys, we use "ObfuscateKeys" so if the client wants to obfuscate, they can activate it.

letzya: What is your reasoning for this change assuming it's the same just the other way around? (I was debating myself with that :) )

letzya:
@tbuchaillot analytics_test.go had the wrong tests, please check again.

analytics/analytics_test.go
```diff
@@ -0,0 +1,93 @@
...
+	//test no filter
+	filter = AnalyticsFilters{}
+	shouldFilter = filter.ShouldFilter(record)
+	if shouldFilter == true {
+		t.Fatal("filter should not be filtering the record")
+	}
+
+}
+
+func TestHasFilter(t *testing.T) {
```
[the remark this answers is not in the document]
letzya: fixed. was the result of copy/paste

tbuchaillot:
@letzya can you `gofmt -s -w` the analytics_filters_test.go file? Besides that, it looks great!

konrad-sol:
When I merged it automatically with master I'm able to reproduce issue - https://tyktech.atlassian.net/browse/TT-1335
@tbuchaillot could you re-merge it with master, so we would be sure that this will not introduce this problem again?

config.go
```diff
@@ -36,6 +41,8 @@ type TykPumpConfiguration struct {
 	HealthCheckEndpointName string                     `json:"health_check_endpoint_name"`
 	HealthCheckEndpointPort int                        `json:"health_check_endpoint_port"`
 	OmitDetailedRecording   bool                       `json:"omit_detailed_recording"`
+	base64DecodeRawData			bool											 `json:"base64_decode_raw_data"`
+	obfuscateAuthHeader		  ObfuscateAuthHeader
```
tbuchaillot, line 45: @letzya I think we should change this or add a json tag in the `obfuscateAuthHeader` field.

If we do
```
base64DecodeRawData			bool											 `json:"base64_decode_raw_data"`
ObfuscateAuthHeader 
```
We could have `obfuscate_keys` and `auth_header_name` at the root config level.

analytics/analytics_test.go
```diff
@@ -0,0 +1,183 @@
+package analytics
+
+import (
+	"encoding/base64"
+	"testing"
+)
+
+func TestObfuscateKeys(t *testing.T) {
```
tbuchaillot, line 8: @letzya most of these tests are not passing, can you help me to understand why?

tbuchaillot (changes requested):

PR #283, opened 2020-09-25
TT-695 Mongo pumps Timeout

tbuchaillot:
## Description

Setting the session timeout for mongo aggregate and selective based on the context.  
This way it could be settable by config.

## Related Issue

https://tyktech.atlassian.net/browse/TT-695

## Motivation and Context

https://tyktech.atlassian.net/browse/TT-695

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/mongo_aggregate.go
+++ b/pumps/mongo_aggregate.go
@@ -227,6 +227,16 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{}) 
 		// calculate aggregates
 		analyticsPerOrg := analytics.AggregateData(data, m.dbConf.TrackAllPaths, m.dbConf.IgnoreTagPrefixList, m.dbConf.StoreAnalyticsPerMinute)
 
+		timeout, deadlineSet := ctx.Deadline()
+		if deadlineSet{
+			duration := time.Duration(timeout.Second()) * time.Second
+			m.dbSession.SetSocketTimeout( duration)
+			m.dbSession.SetSyncTimeout(duration)
+		}
+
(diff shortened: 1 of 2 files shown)


tbuchaillot:
@gernest @matiasinsaurralde any suggestion on how to handle this in the [normal mongo pump](https://github.com/TykTechnologies/tyk-pump/blob/7b283d8f5df0e8f4a32aae2ac852c4da525e2b53/pumps/mongo.go#L455)? 
IMO it's pretty risky on how we spam go routines and copy sessions in each one.

gernest (approved):
Looks good to me. Regarding https://github.com/TykTechnologies/tyk-pump/blob/7b283d8f5df0e8f4a32aae2ac852c4da525e2b53/pumps/mongo.go#L455

I don't think we need that goroutine. because `WriteData` is called in a separate goroutine.

tbuchaillot:
> Looks good to me. Regarding
> 
> https://github.com/TykTechnologies/tyk-pump/blob/7b283d8f5df0e8f4a32aae2ac852c4da525e2b53/pumps/mongo.go#L455
> 
> I don't think we need that goroutine. because `WriteData` is called in a separate goroutine.

@gernest Yup, but I'm afraid of that slowing other pumps when purgeLoop takes a big chunk of data since it would mean waiting for a lot of writings.

gernest:
@tbuchaillot can't you maybe batch them? I see we are doing `analyticsCollection.Insert(dataSet...)` . collect `dataSet` and then do a single insert ?.

PR #292, opened 2020-11-18
Update DogStatsD to support customizable tags

diegobernardes:
There is no option to customize DogStatsD and this may have a negative impact on the billing with tags like `path` that are sent by default.

## Description
This PR adds an option to configure the tags pushed to DogStatsD.

## Motivation and Context
Decrease the cost of Datadog.

## How This Has Been Tested
There are no automated tests for the `dogstatsd.go` file. I did some manual tests and the code works.

## Types of changes

## Checklist
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] Check your code additions will not fail linting checks:
  - [X] `go fmt -s`
  - [X] `go vet`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
+/.vscode
 .vagrant/
 build/
 build_tools/
--- a/README.md
+++ b/README.md
@@ -47,9 +47,7 @@ Create a `pump.conf` file:
   "pumps": {
     "dummy": {
       "type": "dummy",
-      "meta": {
-        
-      }
+      "meta": {}
     },
     "mongo": {
       "type": "mongo",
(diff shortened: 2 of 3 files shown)


matiasinsaurralde:
Thanks, we'll review it!

tbuchaillot:
@diegobernardes thanks for this PR! 
I think it would be better if we maintain the current behavior by default and if `s.conf.Tags > 0` we use this logic to maintain backward compatibility.

diegobernardes:
@tbuchaillot, done. But I really think that the default values, in this case, are bad. This is the command that is present at the `WriteData` function:
```
From DataDog website:
Tags shouldn’t originate from unbounded sources, such as EPOCH timestamps, user IDs, or request IDs. Doing
so may infinitely increase the number of metrics for your organization and impact your billing.

As such, we have significantly limited the available metrics which gets sent to datadog.
```

And that's exactly the problem, we're sending unbounded information by default with the `path` tag. If the tyk instance is a low volume one, it's ok, but with a high volume instance, it's simply inviable to use datadog due to the cost.

buger:
#Let’s then update default config in docs/readme, and new users will get the change and old will be unaffected

diegobernardes:
@buger, no problem. Readme updated.

tbuchaillot (approved):

konrad-sol:
Tested on PR branch -> QA status OK

PR #303, opened 2020-12-29
[TT-1295] Current log messages are confusing if there was a network error

letzya:
fix https://tyktech.atlassian.net/browse/TT-1295

## Description

We print a message that we wrote X number of records to mongo but actually, we print it before we attempt to write and not after, so in case of problems writing to mongo (network issues, delays, etc. ) we ending up dropping these records completely so I thought we should be clearer about the message - before writing we have a debug level of "attempt to write" and after the write has finished successfully we write an info level message.

I also added the collection name to make it easier to understand errors. 

## Motivation and Context

Per description - to make it easier for our clients to debug issues in tyk-pump, mongo etc.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


tbuchaillot:
@letzya, don't you think it would be better if we "normalize" the logs in the way you did for Kafka / Mongo for all the pumps?

letzya:
@tbuchaillot - do you mean - obfuscate for **APIKey** field for unhashed keys? could do but would first check if clients want it.

letzya:
@tbuchaillot that's a great suggestion!

letzya:
closing in favour of @tbuchaillot 's fabulous [PR](https://github.com/TykTechnologies/tyk-pump/pull/319)

PR #309, opened 2021-01-29
 Add capability to obfuscate keys for all records

letzya:
## Description

While filtering the record I thought it's good timing to obfuscate the key. For every record, I check if it needs to be filtered out and then obfuscate the key.
There's a global config field to turn this on - obfuscate_keys
Added TestObfuscateKeys just for that obfuscation function.
Added tests for filter+obfuscation of keys in a record

~Important: This PR is only useful if keys are not hashed, otherwise there's no real reason to obfuscate the key but it works for both cases.~
About hashed keys (that is set in the GW):
- In the case of detailed recording, it doesn't matter since we record the request as-is, and as such, you can see the authorization header in the raw request.
- If there is no detailed recording set, then APIKey field in the analytic record has the key so customers that don't hash keys might want to obfuscate this field to avoid leak when data go outside the tyk management platform.

## Motivation and Context

When recording the full request body, the authorization header gets recorded as well. From obvious reasons, sometimes this is problematic for clients. For example, you don't want the people that check the analytics to have access to user's keys.
Important to note that it's not enough, that it'll be better to add such fix in the source, i.e. the gateway itself should remove this header before saving the analytic record.

## How This Has Been Tested

- Wrote unit tests to test `ObfuscateKey` function
- Wrote tests to test `filterData` that call `ObfuscateKey` - tests with data to filter and without.

How to test:
1. Hashed keys - `"hash_keys": true`:
    1.1. "detailed_recording" on - you can set it in the key. 
         `"obfuscate_key": true` in pump.conf and make API call.
         **Result**: the analytic record of the call should have "****wxyz" instead of the **actual** key in the raw request field and APIKey field should look similar (similar since it's a **hashed** key).
   1.2. "detailed_recording" off - 
         `"obfuscate_key": true` in pump.conf and make API call.
         **Result**: the analytic record of the call should have "****abcd" instead of the **hashed** key in the APIKey field.
2. Cleartext keys -`"hash_keys": false`:
 2.1. "detailed_recording" on - you can set it in the key. 
         `"obfuscate_key": true` in pump.conf and make API call.
         **Result**: the analytic record of the call should have "****wxyz" instead of the **actual** key in the raw request field and "****wxyz" in the APIKey field.
   2.2. "detailed_recording" off - 
         `"obfuscate_key": true` in pump.conf and make API call.
         **Result**: the analytic record of the call should have "****wxyz" instead of the **actual** key in the APIKey field.
3. Test the above with filter data, for instance, set to filter api1 and call api1 and api2 and verify it still works as expected.
```
"filters":{
  "api_ids":[<API-ID>],
```

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -2,6 +2,7 @@ package analytics
 
 import (
 	"encoding/base64"
+	"fmt"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


analytics/analytics.go
```diff
@@ -121,85 +122,76 @@ func (a *AnalyticsRecord) GetLineValues() []string {
 }
 
 //change name - obfuscateAndDecode request
-func (a *AnalyticsRecord) ObfuscateKey(authHeaderName string, decode bool) {
```
letzya, line 124: @tbuchaillot even those I reverted this commit, it's still showing up in the diff

config.go
```diff
@@ -41,8 +41,7 @@ type TykPumpConfiguration struct {
 	HealthCheckEndpointName string                     `json:"health_check_endpoint_name"`
 	HealthCheckEndpointPort int                        `json:"health_check_endpoint_port"`
 	OmitDetailedRecording   bool                       `json:"omit_detailed_recording"`
-	base64DecodeRawData			bool											 `json:"base64_decode_raw_data"`
-	obfuscateAuthHeader		  ObfuscateAuthHeader
+	obfuscateAuthHeader     ObfuscateAuthHeader        `json:"obfuscate_auth_header"`
```
letzya, line 44: @tbuchaillot `ObfuscateAuthHeader` struct is in master even though when rebasing - it's completely disappearing

tbuchaillot (changes requested):
@letzya could you remove prints and unnecessary debug logs?

letzya:
@tbuchaillot removed debug. Also please check the diff in analytics/analytics_test.go

tbuchaillot (approved):
LGTM

konrad-sol:
tested on hashed and not-hashed systems -> works fine

PR #313, opened 2021-02-09
[TD-10] Goreleaser for more repos

alephnull:
# Description
These related PRs add a goreleaser based workflow for the following distribution channels
- Ubuntu and Debian
- RHEL/OL
- tarballs
- docker hub
- devenv ECR
- AWS mktplace
for the following repos
- tyk
- tyk-analytics
- tyk-pump
- tyk-sink

## Pending
- [ ] tykio/golang-cross
- [ ] go version needed for gw, dash, pump and sink
- [ ] AMIs for tyk and pump

# Testing
The status of the integration workflows can be inspected in this PR itself in the _Checks_ tab.

# Merging
Once this PR is merged, the following will be deprecated
- Buddy actions that build packages
- Buddy actions that build AMIs
- int-image.yml


jlucktay (changes requested):
There's a [red light on a Docker build](https://github.com/TykTechnologies/tyk-pump/pull/313/checks?check_run_id=1935516346):
`COPY failed: file not found in build context or excluded by .dockerignore: stat install/data/tyk-pump.docker.conf: file does not exist`

Should this be `install/data/pump.template.conf` or another file?

alephnull:
> Should this be `install/data/pump.template.conf` or another file?

Fixed in a141055

.goreleaser.yml
```diff
@@ -0,0 +1,188 @@
...
+      - std-arm64
+    formats:
+      - deb
+      - rpm
+    contents:
+      - src: "README.md"
+        dst: "/opt/share/docs/tyk-pump/README.md"
+      - src: "/opt/tyk-pump"
+        dst: "/opt/tyk-pump"
+        type: "symlink"
```
excieve, line 113: Hmm... how does this symlink work? :)

Dockerfile.std
```diff
@@ -0,0 +1,23 @@
...
+FROM debian:buster-20210208-slim
+RUN apt-get update \
+ && apt-get dist-upgrade -y ca-certificates \
+ && apt-get autoremove -y
+COPY *.deb /
+RUN dpkg -i /*.deb && rm /*.deb
+
+ARG PORTS
+EXPOSE $PORTS
+VOLUME [ "/opt/tyk-pump" ]
```
excieve, line 17: Same about the volume exposure as in the other PRs of this set.

install/data/tyk-pump.docker.conf
```diff
@@ -0,0 +1,74 @@
...
+    "mongo": {
+      "type": "mongo",
+      "meta": {
+        "collection_name": "tyk_analytics",
+        "mongo_url": "mongodb://localhost/tyk_analytics",
+        "collection_cap_max_size_bytes": 1048576,
+        "collection_cap_enable": true
+      }
+    },
+    "kafka": {
```
excieve, line 26: I don't think default conf needs kafka, prom, etc. samples in the same file. Mongo should be sufficient for the default use case.

excieve (commented):
Same comments as others in this set. Also the package symlinks the dir to self, which I think creates a recursive link inside a directory — that's a confusing artefact :)

install/data/tyk-pump.docker.conf
alephnull: Done.

.goreleaser.yml
alephnull: Thanks!

Dockerfile.std
alephnull: Done.

PR #315, opened 2021-02-16
Revert " Add capability to obfuscate keys for all records"

letzya:
Reverts TykTechnologies/tyk-pump#309

This feature has a bug and is not working. We are reverting this and reconsider the options.


matiasinsaurralde:
Do we have any details on the bug?

letzya:
Yes- it's doing _strings.Split_ on the api key, but if it's hashed - it will never find it.

tbuchaillot (approved):
LGTM

PR #317, opened 2021-02-25
Splunk Pump - Customizable analytics fields

mbelchev:
## Description

The purpose of this pull request is to extend the Tyk Splunk Pump with new functionality. This new functionality is related to the analytics data that is being sent to Splunk.

With the changes in this pull request, now we will be able to easily configure which Tyk analytic fields to be sent to Splunk.

Also, the API key obfuscation has been improved so you can now set how long the partially shown API key be.

## Related Issue

[#12059](https://support.tyk.io/hc/en-gb/requests/12059)

## Motivation and Context

Currently, each analytics record has around 30 items that can be passed to any data collector. But the Splunk Pump is sending only half of them. As a result, the received events lack information and are not optimal for troubleshooting.

## How This Has Been Tested

The suggested feature has been tested in a real and active environment in our company. For more details, we can discuss in the related ticket.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/pumps/splunk.go
+++ b/pumps/splunk.go
@@ -145,18 +145,12 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
 	}).Info("Writing ", len(data), " records")
 	for _, v := range data {
 		decoded := v.(analytics.AnalyticsRecord)
-		apiKey := decoded.APIKey
-		if p.config.ObfuscateAPIKeys {
-			if len(apiKey) > 4 {
-				apiKey = "****" + apiKey[len(apiKey)-4:]
-			}
-		}
 
 		event := map[string]interface{}{
 			"method":        decoded.Method,
 			"path":          decoded.Path,
 			"response_code": decoded.ResponseCode,
-			"api_key":       apiKey,
+			"api_key":       decoded.APIKey,
 			"time_stamp":    decoded.TimeStamp,
 			"api_version":   decoded.APIVersion,
 			"api_name":      decoded.APIName,
--- a/pumps/splunk.go
+++ b/pumps/splunk.go
@@ -96,14 +96,15 @@ type SplunkPump struct {
 
 // SplunkPumpConfig contains the driver configuration parameters.
 type SplunkPumpConfig struct {
-	CollectorToken        string   `mapstructure:"collector_token"`
-	CollectorURL          string   `mapstructure:"collector_url"`
-	SSLInsecureSkipVerify bool     `mapstructure:"ssl_insecure_skip_verify"`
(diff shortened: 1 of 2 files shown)


pumps/splunk.go
```diff
@@ -144,29 +146,73 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
...
+		if len(p.config.Fields) > 0 {
+			// Loop through all fields set in the pump config
+			for _, field := range p.config.Fields {
+				// Skip the next actions in case the configured field doesn't exist
+				if _, ok := mapping[field]; !ok {
+					continue
+				}
+
+				// Check if the field is "api_key" and the obfuscation is configured
+				if field == "api_key" && p.config.ObfuscateAPIKeys {
```
tbuchaillot, line 185: @mbelchev, in this way, we're obfuscating the key only if `api_key` is in the `fields` array. I would do it before creating the `mapping` map, so both config options are independent of each other. 
What do you think?

tbuchaillot (changes requested):

pumps/splunk.go
mbelchev: @tbuchaillot, I was thinking about it but also considered that there might be a future improvement that will give the end-user the option to customize all fields on the fly with some kind of hooks. But for now, I agree with you and the change will be applied. Thanks for your feedback!

tbuchaillot (approved):
LGTM

matiasinsaurralde (approved):

tbuchaillot:
@mbelchev merged :) Thanks for the contribution!

pumps/splunk.go
```diff
@@ -145,28 +147,72 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
...
-		if p.config.ObfuscateAPIKeys {
-			if len(apiKey) > 4 {
-				apiKey = "****" + apiKey[len(apiKey)-4:]
-			}
+
+		// Check if the APIKey obfuscation is configured and its doable
+		if p.config.ObfuscateAPIKeys && len(apiKey) > p.config.ObfuscateAPIKeysLength {
+			// Obfuscate the APIKey, starting with 4 asterics and followed by last N chars (configured separately) of the APIKey
+			// The default value of the length is 0 so unless another number is configured, the APIKey will be fully hidden
+			apiKey = "****" + apiKey[len(apiKey)-p.config.ObfuscateAPIKeysLength:]
```
letzya, line 155: Why do you need to obfuscate this field? with auth token this field supposed to have hashed value (unless you set the GW to not hash keys)

PR #320, opened 2021-03-03
[TT-1574] Multiple analytics keys

tbuchaillot:
## Description

This PR adds support for multiple analytic keys and is backward compatible with the old analytic key (`tyk-system-analytic`).

It just basically do the following process:
Iterate from -1 to 9:
- If the iteration is -1 (the first one), we're going to look for the old analytic key (`tyk-system-analytic`)
- If the iteration is from 0 to 9, we're going to look for the new divided key (`tyk-system-analytic_X` where X is the iteration number)
- Read from the Redis the key 
- Process data
- Jump to the next key

This allows us to:
- Loss fewer analytics records if a pump fails since it will only lose the records from one of the keys.
- If the Redis cluster is enabled, read the analytics across multiple nodes. 

## Related Issue

https://tyktech.atlassian.net/browse/TT-1574
## Motivation and Context

Reduce redis overload and have a better analytics distribution across the cluster. Also to have better fault tolerance in the pump.

## How This Has Been Tested

- Set enable_multiple_analytics_keys:true in the pump and analytics_config.enable_multiple_analytics_keys:true in the gateway.
- Using redis cluster (https://github.com/Grokzen/docker-redis-cluster) or/and standalone Redis.
- Setup any pump (CSV and dummy in my tests).
- Send 100 requests to a gw API.
- Your pumps should process 100 analytics records.

If you keep sending requests and turn off the pump, you should see in Redis multiple analytics keys with the following format: tyk-system-analytics_X where X goes from 0 to 9. And each key should have different records inside.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -163,37 +164,48 @@ func StartPurgeLoop(secInterval int, chunkSize int64, expire time.Duration, omit
...
-		if len(AnalyticsValues) > 0 {
-			startTime := time.Now()
+		iterationsPerLoop := 1
+		if SystemConfig.EnableMultipleAnalyticsKeys {
+			iterationsPerLoop = 10
+		}
+		for i := 0; i < iterationsPerLoop; i++ {
+			analyticsKeyName := storage.ANALYTICS_KEYNAME
+			if SystemConfig.EnableMultipleAnalyticsKeys {
+				analyticsKeyName += "_" + fmt.Sprint(i)
```
buger, line 174: I suggest to make it backward compatible, and read "old" key without prefix too. 
By doing so it will be safe upgrade pump first, and it will continue working with old gateways.

tbuchaillot: @buger if `enable_multiple_analytics_keys` is false, it's backward compatible since it will only loop once https://github.com/TykTechnologies/tyk-pump/blob/8ded20fc66eafdef07f22089330fe091acdbe81b/main.go#L168.

main.go
```diff
@@ -162,40 +163,48 @@ func initialisePumps() {
...
-					"prefix": mainPrefix,
-				}).Debug("Decoded Record: ", decoded)
-				if err != nil {
+		for i := -1; i < 10; i++ {
+			var analyticsKeyName string
+			if i == -1 {
+				//if it's the first iteration, we look for tyk-system-analytics to maintain backwards compatibility or if analytics_config.enable_multiple_analytics_keys is disabled in the gateway
+				analyticsKeyName = storage.ANALYTICS_KEYNAME
+			} else {
+				analyticsKeyName = storage.ANALYTICS_KEYNAME + "_" + fmt.Sprint(i)
```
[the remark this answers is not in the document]
tbuchaillot: done @sredxny

sredxny (approved):
lgmt

PR #346, opened 2021-05-05
[TT-2426] Stdout json pump

joshblakeley:
## Description

Enables  JSON or Text logging of analytics records to stdout for scraping from containers etc by logging agents such as Datadog/Logly/Logstash etc

Users can configure if they require JSON or Text formatted logs and also the json field name of the analytics record 

Configuration looks like this:

```
"pumps": {
    "stdout":{
      "type": "stdout",
      "meta" : {
      "log_field_name": "tyk-analytics-record",
      "format": "json"
      }
    }
  },
```

Also added a general configuration for all the logs.
`log_level` - Set the logger details for tyk-pump. The posible values are: `info`,`debug`,`error` and `warn`. By default, the log level is `info`. 

`log_format` - Set the logger format. The possible values are: `text` and `json`. By default, the log format is `text`.

## Motivation and Context

Clients require simple methods to extract raw request analytics to dedicated logging platforms. In containerised envs scraping the containers stdout is a simple and effective way to achieve this.

Screenshot 2021-05-05 at 08 47 05

## Types of changes

## Checklist

- [X] My change requires a change to the documentation.
  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/stdout.go
```diff
@@ -0,0 +1,100 @@
...
+
+	s.conf = &StdOutConf{}
+	err := mapstructure.Decode(config, &s.conf)
+
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"prefix": statsdPrefix,
+		}).Fatal("Failed to decode configuration: ", err)
+	}
+
```
tbuchaillot, line 42: need to parse env vars here. Take a look at the master csv pump https://github.com/TykTechnologies/tyk-pump/blob/c9d9b92ecea143423e40acdfa5e2d3d71b5bb63d/pumps/csv.go#L52

pumps/stdout.go
```diff
@@ -0,0 +1,100 @@
...
+	log = logger.GetLogger()
```
tbuchaillot, line 43: Try to initialise the logger with a prefix like we do in the other pumps. https://github.com/TykTechnologies/tyk-pump/blob/c9d9b92ecea143423e40acdfa5e2d3d71b5bb63d/pumps/csv.go#L45

tbuchaillot (changes requested):
@joshblakeley make sure you pull the latest `master` changes. 

Now the pump interface requires each pump to have `GetEnvPrefix()` method.

Also, it would be great if we add the pump / usage  at README.md

pumps/stdout.go
joshblakeley: I actually only get the global logger to override to JSON output if needed

joshblakeley:
@tbuchaillot done and done

pumps/stdout.go
```diff
@@ -0,0 +1,109 @@
...
+	err := mapstructure.Decode(config, &s.conf)
+
+	if err != nil {
+		s.log.Fatal("Failed to decode configuration: ", err)
+	}
+
+	processPumpEnvVars(s, s.log, s.conf, stdOutDefaultENV)
+
+	if s.conf.Format == "json" {
+		log.Formatter = &logrus.JSONFormatter{}
```
tbuchaillot, line 53: @joshblakeley  is this going to change the format of all our logs? I mean, if we configure 2 pumps (one being stdout pump), all the pumps are going to print in json format?

tbuchaillot (changes requested):

pumps/stdout.go
joshblakeley: Its configured on each pump. My understanding is with pump you need to have them all with the same config unless you have filters anyway because they greedily compete for the same redis key.

tbuchaillot: It seems that when you modify the formatter from `log` (shared between pumps), it modifies the format for every pump
image

Also, the pumps don't compete for the same key, they all process the same key concurrently.

tbuchaillot:
@joshblakeley let me know what you think about the changes.

pumps/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+				"user_agent":      decoded.UserAgent,
+			}
+
+			if s.conf.Format == "json" {
+				formatter := &logrus.JSONFormatter{}
+
+				entry := log.WithField(s.conf.LogFieldName, message)
+				entry.Level = logrus.InfoLevel
+				data, _ := formatter.Format(entry)
+				fmt.Print(string(data))
```
joshblakeley, line 107: @tbuchaillot is this just debugging that you need to remove

joshblakeley (commented):
Small debugging print to remove but otherwise looks good

pumps/stdout.go
tbuchaillot: @joshblakeley that's the line that prints the json in the terminal. Why do we need to remove it?

tbuchaillot: @joshblakeley that's the line that prints the analytic record in json format in the console. Why do we need to remove it?

joshblakeley: Its weird we hand it to fmt to handle - why not logrus directly?

lopeg:
Awesome PR!
Looking forward to using it
thank you @joshblakeley

pumps/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+				"org_id":          decoded.OrgID,
+				"oauth_id":        decoded.OauthID,
+				"raw_request":     decoded.RawRequest,
+				"request_time_ms": decoded.RequestTime,
+				"raw_response":    decoded.RawResponse,
+				"ip_address":      decoded.IPAddress,
+				"host":            decoded.Host,
+				"content_length":  decoded.ContentLength,
+				"user_agent":      decoded.UserAgent,
+			}
```
joshblakeley, line 99: I can see a few more fields missing we should definitely add - basically lets get the whole analytics record in there! 

Things like tags are super important

https://github.com/TykTechnologies/tyk-pump/blob/c9d9b92ecea143423e40acdfa5e2d3d71b5bb63d/analytics/analytics.go#L28

joshblakeley (commented):
One more note about expanding the analytics record. Apart from that good to go if you are happy with there being a mix of JSON and TEXT in the output

Also please update the docs in the PR description and README based on your changes :)

pumps/stdout.go
tbuchaillot: @joshblakeley because logrus adds a lot of "info" to the message and it's not in JSON format. If we modify logrus format, we modify it at a general logger level and that's something that we don't want (at least here).

What's the weird part of  `fmt` handling it?

pumps/stdout.go
tbuchaillot:
```json 
{"level":"info","msg":"","time":"0001-01-01T00:00:00Z","tyk-analytics-record":{"method":"PUT","host":"","path":"/widget/unconvergent","raw_path":"/widget/unconvergent","content_length":609,"user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/602.4.8 (KHTML, like Gecko) Version/10.0.3 Safari/602.4.8","day":5,"month":5,"year":2021,"hour":22,"response_code":200,"api_key":"588b52eab275ff0001cc7476ff2be18c970949ac81a30582abf9c445","timestamp":"2021-05-05T22:42:07.133794-03:00","api_version":"Default","api_name":"Foo Bar","api_id":"a47f35f411f14c837b99e99519e2432c","org_id":"588b52eab275ff0001cc7476","oauth_id":"","request_time":3,"raw_request":"Qk9EWSBEQVRB","raw_response":"UkVTUE9OU0UgREFUQQ==","ip_address":"118.93.55.103","geo":{"country":{"iso_code":""},"city":{"geoname_id":0,"names":null},"location":{"latitude":0,"longitude":0,"time_zone":""}},"network_stats":{"open_connections":0,"closed_connections":0,"bytes_in":0,"bytes_out":0},"latency":{"total":0,"upstream":0},"tags":["orgid-588b52eab275ff0001cc7476","apiid-a47f35f411f14c837b99e99519e2432c"],"alias":"","track_path":true,"expireAt":"2022-05-27T22:42:07.625944-03:00"}}
```

Something like that would do the trick? I think I'm not missing any field 👀

tbuchaillot:
@joshblakeley updated :) let me know what do you think

tbuchaillot (approved):

matiasinsaurralde (approved):

PR #359, opened 2021-06-11
releng: [TD-388,TD-413] fixes and upgrade tests

alephnull:
## Description
- remove cloudsmith multi-arch manifests
- automatic upgrade tests on push

## Related Issue
https://tyktech.atlassian.net/browse/TD-351
https://tyktech.atlassian.net/browse/TD-374
https://tyktech.atlassian.net/browse/TD-373

## How This Has Been Tested
See the status in the _Checks_ tab.

# Merging
This commit needs to be cherry-picked to release-1.8. If the tykbot cherry pick fails, create a fresh PR for release-1.8 instead of a rebase.


tbuchaillot (approved):

.github/workflows/release.yml
```diff
@@ -235,25 +241,33 @@ jobs:
...
 
       - name: generate dockerfile
         run: |
           echo 'FROM ${{ matrix.distro }}
-          COPY tyk-pump*_amd64.deb /tyk-pump.deb
+          ARG TARGETARCH
+          COPY tyk-pump*_${TARGETARCH}.deb /tyk-pump.deb
+          RUN apt-get update && apt-get install -y curl
+
+          RUN curl -fsSL https://packagecloud.io/install/repositories/tyk/tyk-pump/script.deb.sh | bash && apt-get install -y tyk-pump=1.2.0
```
buger, line 255: Is it ok that 1.2.0 is hard coded here?

alephnull: So this is the version that we will upgrade from. It is chosen as the last stable release before goreleaser landed. It is defined for all the repos at https://github.com/TykTechnologies/tyk-ci/blob/master/wf-gen/header.m4#L12

.github/workflows/release.yml
```diff
@@ -235,25 +241,33 @@ jobs:
...
           RUN dpkg -i /tyk-pump.deb && /opt/tyk-pump/tyk-pump --conf=/opt/tyk-pump/tyk-pump.conf &' > Dockerfile
```
buger, line 256: Does it actually checks if pump run successfully? 
Can we have some simple "custom" tyk-pump.conf, with "CSV" pump or similar, and after running pump, check its STDOUT to ensure that old config was loaded (check for CSV keyword for example, or lack of mongo keyword)

alephnull: Yes, this will be a better test. I do not have a known good config file nor am I clear on how it can be tested. @ilijabojanovic could you help me with a better test case?

.github/workflows/release.yml
buger: So with every new major release we should not forget to update it? If so, just ensure that it is included in some checklist.

.github/workflows/release.yml
buger: Here is the config you can use for pump:

```
{
    "analytics_storage_type": "redis",
    "analytics_storage_config": {
        "type": "redis",
        "host": "localhost",
        "port": 6379,
        "username": "",
        "password": "",
        "database": 0,
        "optimisation_max_idle": 500,
        "optimisation_max_active": 1000
    },
    "purge_delay": 5,
    "dont_purge_uptime_data": true,
    "pumps": {
      "csv": {
        "type": "csv",
        "meta": {
          "csv_dir": "./"
        }
      }
    }
}
```

In STDOUT you should check for `Init Pump: CSV Pump` message.

buger: Only redis is dependency, mongo is not required.

.github/workflows/release.yml
alephnull: I do not think it is _necessary_ to update this version for every update. Another option is to have an array of old versions and test every version.

buger: Yeah, array does make sense, to spot issues between bigger upgrades.

alephnull: As for adding it to the checklist, would you please point me to the checklist that needs to be added to or the person who runs the checklist?

.github/workflows/release.yml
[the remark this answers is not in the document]
alephnull: This is actually a very simple test that only checks that we have the correct architecture. All it says that the program can be invoked. @buger suggests a more robust test above which I need your help with.

I'll update the config file anyways.

.github/workflows/release.yml
alephnull: Can we push making it array for the future? Right now we don't have an array.

.github/workflows/release.yml
alephnull: This sort of test is not easy to "templatetise" into https://github.com/TykTechnologies/tyk-ci/blob/master/wf-gen/release/install.m4#L39 for all the repos. I will think about how this could be best implemented.

Alternatively, a pump is running in the dev envs and is updated for every commit. It would make sense to run a wider set of tests against this.

buger: Unlikely you can have "generalized" template which fits all repos. But you can add different actions based on repo name. E.g. different upgrade tests.

alephnull: Indeed. This is a problem that needs to be solved for the plugin tests and implementing it is more complex which is why I have not done it in this PR. Do you want to keep this PR open until I work all of this out?

PR #364, opened 2021-06-29
[TT-2117] Feature/uptime sql

tbuchaillot:
## Description

Adding uptime SQL pump.

Now, you can configure a `uptime_type` in the `uptime_pump_config`, and depending on that, you can set the configuration fields. By default, the `uptime_type` is `mongo`. For SQL Uptime pumps, the `uptime_type` is `sql`.

In order to set uptime SQL pump:
```
"uptime_pump_config": {
    "uptime_type":"sql",
    "type":"sqlite",
    "connection_string":"/Users/tomuz/test.db",
    "table_sharding":false,
  },
```

The uptime SQL Pump stores the data in an aggregated way, similar to `sql_aggregate`. It supports `table_sharding` and by default, the **aggregations are per hour**.

## Related Issue

https://tyktech.atlassian.net/browse/TT-2117

## Motivation and Context

https://tyktech.atlassian.net/browse/TT-2117

## How This Has Been Tested

Added unit tests and tested manually with `sqlite` configured in the dashboard analytics storage.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -24,7 +24,7 @@ type Latency struct {
 	Upstream int64 `json:"upstream"`
 }
 
-const SQLTABLE = "tyk_analytics"
+const SQLTable = "tyk_analytics"
 
 // AnalyticsRecord encodes the details of a request
 type AnalyticsRecord struct {
@@ -60,7 +60,7 @@ type AnalyticsRecord struct {
 }
 
 func (ar *AnalyticsRecord) TableName() string {
-	return SQLTABLE
+	return SQLTable
 }
 
 type GeoData struct {
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -24,6 +24,8 @@ type Latency struct {
 	Upstream int64 `json:"upstream"`
 }
 
+const SQLTABLE = "tyk_analytics"
+
 // AnalyticsRecord encodes the details of a request
 type AnalyticsRecord struct {
 	Method        string       `json:"method" gorm:"column:method"`
@@ -58,7 +60,7 @@ type AnalyticsRecord struct {
 }
 
 func (ar *AnalyticsRecord) TableName() string {
-	return "tyk_analytics"
+	return SQLTABLE
 }
 
 type GeoData struct {
--- a/analytics/uptime_data.go
+++ b/analytics/uptime_data.go
@@ -1,20 +1,326 @@
 package analytics
 
-import "time"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/fatih/structs"
+	"gorm.io/gorm"
+)
 
 type UptimeReportData struct {
-	URL          string
-	RequestTime  int64
-	ResponseCode int
-	TCPError     bool
-	ServerError  bool
-	Day          int
-	Month        time.Month
-	Year         int
-	Hour         int
-	Minute       int
-	TimeStamp    time.Time
-	ExpireAt     time.Time `bson:"expireAt" json:"expireAt"`
-	APIID        string
-	OrgID        string
+	URL          string `json:"url" gorm:"column:url"`
+	RequestTime  int64 `json:"request_time" gorm:"column:requesttime"`
+	ResponseCode int `json:"response_code" gorm:"column:responsecode;index"`
+	TCPError     bool `json:"tcp_error" gorm:"tcperror"`
+	ServerError  bool `json:"server_error" gorm:"servererror"`
+	Day          int `json:"day" gorm:"day"`
+	Month        time.Month `json:"month" gorm:"month"`
+	Year         int `json:"year" gorm:"year"`
+	Hour         int `json:"hour" gorm:"hour"`
+	Minute       int `json:"minute" sql:"-"`
+	TimeStamp    time.Time `json:"timestamp" gorm:"column:timestamp;index"`
+	ExpireAt     time.Time `bson:"expireAt" json:"expireAt" gorm:"expireAt"`
+	APIID        string `json:"api_id" gorm:"column:apiid;index"`
+	OrgID        string `json:"org_id" gorm:"column:orgid;index"`
+}
+
+type UptimeReportAggregateSQL struct {
+	ID      string  `gorm:"primaryKey"`
+
+	Counter `json:"counter" gorm:"embedded"`
+
+
+	TimeStamp      int64  `json:"timestamp" gorm:"index:dimension, priority:1"`
+	OrgID          string `json:"org_id" gorm:"index:dimension, priority:2"`
+	Dimension      string `json:"dimension" gorm:"index:dimension, priority:3"`
+	DimensionValue string `json:"dimension_value" gorm:"index:dimension, priority:4"`
+
+	Code1x  int `json:"code_1x"`
+	Code200 int `json:"code_200"`
+	Code201 int `json:"code_201"`
+	Code2x  int `json:"code_2x"`
+	Code301 int `json:"code_301"`
+	Code302 int `json:"code_302"`
+	Code303 int `json:"code_303"`
+	Code304 int `json:"code_304"`
+	Code3x  int `json:"code_3x"`
+	Code400 int `json:"code_400"`
+	Code401 int `json:"code_401"`
+	Code403 int `json:"code_403"`
+	Code404 int `json:"code_404"`
+	Code429 int `json:"code_429"`
+	Code4x  int `json:"code_4x"`
+	Code500 int `json:"code_500"`
+	Code501 int `json:"code_501"`
+	Code502 int `json:"code_502"`
+	Code503 int `json:"code_503"`
+	Code504 int `json:"code_504"`
+	Code5x  int `json:"code_5x"`
+}
+
+
+func (a UptimeReportAggregateSQL) GetAssignments() map[string]interface{} {
+	assignments := make(map[string]interface{})
+
+	fields := structs.Fields(a.Counter)
+	for _, field := range fields {
+		colName := "counter_"+field.Tag("json")
+
+		switch {
+		case strings.Contains(colName, "hits"), strings.Contains(colName, "error"), strings.Contains(colName, "success"):
+			if !field.IsZero(){
+				assignments[colName] = gorm.Expr("`"+colName + "` + "+fmt.Sprint(field.Value()))
+			}
+		case strings.Contains(colName, "total_request_time"):
+			if !field.IsZero() {
+				assignments[colName] = gorm.Expr("`" + colName + "` + " + fmt.Sprint(a.TotalRequestTime))
+			}
+		case strings.Contains(colName, "request_time"):
+			//AVG adding value to another AVG: newAve = ((oldAve*oldNumPoints) + x)/(oldNumPoints+1)
+			if !field.IsZero() {
+				assignments[colName] = gorm.Expr("((`" + colName + "` * `counter_hits`) + " + fmt.Sprint(a.RequestTime) + ")/( `counter_hits` +1)")
+			}
+		}
+
+	}
+
+	return assignments
+}
+
+func (u *UptimeReportAggregate) Dimensions() (dimensions []Dimension) {
+	for key, inc := range u.URL {
+		dimensions = append(dimensions, Dimension{"url", key, inc})
+	}
+
+	for key, inc := range u.Errors {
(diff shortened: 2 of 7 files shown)


analytics/uptime_data.go
```diff
@@ -1,20 +1,333 @@
...
+
+	return agg
+}
+
+func AggregateUptimeData(data []UptimeReportData) map[string]UptimeReportAggregate {
+	analyticsPerOrg := make(map[string]UptimeReportAggregate)
+
+	for _, v := range data {
+		thisV := v
+		orgID := thisV.OrgID
```
vaske, line 205: out of curiosity any reason why you assign `v` to `thisV`? since it's in same scope it's not necesary you can in for signature say 

```
for _, thisV := range data {
```

really minor thing, not merge stopper :)

pumps/sql_test.go
```diff
@@ -119,3 +121,163 @@ func TestSQLWriteDataSharded(t *testing.T) {
...
+
+	table := "tyk_uptime_analytics_" + now.Format("20060102")
+	assert.Equal(t, true, pmp.db.Migrator().HasTable(table))
+
+	var dbRecords []analytics.UptimeReportAggregateSQL
+
+	if err := pmp.db.Table(table).Find(&dbRecords).Error; err != nil {
+		t.Fatal("Error getting analytics records from SQL")
+	}
+	assert.Len(t, dbRecords, 2)
```
furkansenharputlu, line 224: Why is this one `2`? Shouldn't it be `3` because of `keys[0]`, `keys[1]`, `keys[2]`?

pumps/sql_test.go
```diff
@@ -119,3 +121,163 @@ func TestSQLWriteDataSharded(t *testing.T) {
...
+	}
+	defer func() {
+		os.Remove("pmp_test.db")
+	}()
+
+	keys := make([]interface{}, 3)
+	now := time.Now()
+	nowPlus1 := time.Now().Add(2 * time.Hour)
+
+	encoded, _ := msgpack.Marshal(analytics.UptimeReportData{OrgID: "1", URL: "url1", TimeStamp: now})
```
furkansenharputlu, line 143: In this test, there is repeating code. Wouldn't it be nicer if we put them in an inner func?

analytics/uptime_data.go
```diff
@@ -1,20 +1,333 @@
...
+	ID string `gorm:"primaryKey"`
+
+	Counter `json:"counter" gorm:"embedded"`
+
+	TimeStamp      int64  `json:"timestamp" gorm:"index:dimension, priority:1"`
+	OrgID          string `json:"org_id" gorm:"index:dimension, priority:2"`
+	Dimension      string `json:"dimension" gorm:"index:dimension, priority:3"`
+	DimensionValue string `json:"dimension_value" gorm:"index:dimension, priority:4"`
+
+	Code1x  int `json:"code_1x"`
```
furkansenharputlu, line 40: Why do we put them separately ? Instead what if we create a map and get rid of `ProcessStatusCodes` etc.?

pumps/sql.go
```diff
@@ -201,3 +216,92 @@ func (c *SQLPump) WriteData(ctx context.Context, data []interface{}) error {
...
+
+func (c *SQLPump) WriteUptimeData(data []interface{}) {
+	c.log.Debug("Attempting to write ", len(data), " records...")
+
+	var typedData []analytics.UptimeReportData
+
+	batch := 500
+	newBatch := false
+	var firstDate time.Time
+	for i, r := range data {
```
furkansenharputlu, line 228: if `c.SQLConf.TableSharding` is `false`, what is the purpose of this `for`?

pumps/sql.go
```diff
@@ -201,3 +216,92 @@ func (c *SQLPump) WriteData(ctx context.Context, data []interface{}) error {
...
+		}
+		if i == 0 {
+			firstDate = decoded.TimeStamp
+		}
+
+		typedData = append(typedData, decoded)
+
+		if c.SQLConf.TableSharding && !newBatch && firstDate.Format("20060102") < decoded.TimeStamp.Format("20060102") {
+			batch = i
+			newBatch = true
```
furkansenharputlu, line 245: After the `newBatch` is `true`, why don't we break the for?

pumps/sql.go
```diff
@@ -201,3 +216,92 @@ func (c *SQLPump) WriteData(ctx context.Context, data []interface{}) error {
...
+			c.db.Table(table).AutoMigrate(&analytics.UptimeReportAggregateSQL{})
+		}
+
+		table = "tyk_uptime_analytics_" + typedData[len(typedData)-1].TimeStamp.Format("20060102")
+		if !c.db.Migrator().HasTable(table) {
+			c.db.Table(table).AutoMigrate(&analytics.UptimeReportAggregateSQL{})
+		}
+	}
+
+	for i := 0; i < len(typedData); i += batch {
```
furkansenharputlu, line 262: Are we sure that the `typedData` can only include `2` days?

pumps/sql_test.go
tbuchaillot: @furkansenharputlu nope. Given the uptime data is saved aggregated, here it should 2 have 2 dimensions: `url` and `empty-total`.

pumps/sql_test.go
tbuchaillot: @furkansenharputlu do you mean the encoding part?

pumps/sql.go
tbuchaillot: we need to transform the data from `[]interface` to `[]analytics.UptimeReportData`,  decoding all the items on data.

analytics/uptime_data.go
tbuchaillot: solved! thanks

analytics/uptime_data.go
tbuchaillot: @furkansenharputlu I don't follow you here :(

pumps/sql.go
```diff
@@ -201,3 +216,92 @@ func (c *SQLPump) WriteData(ctx context.Context, data []interface{}) error {
...
+
+		typedData = append(typedData, decoded)
+
+		if c.SQLConf.TableSharding && !newBatch && firstDate.Format("20060102") < decoded.TimeStamp.Format("20060102") {
+			batch = i
+			newBatch = true
+		}
+	}
+
+	if c.SQLConf.TableSharding && len(typedData) > 0 {
```
furkansenharputlu, line 249: As we discussed in pair programming, we can get rid of this if statement.

tbuchaillot: solved!

analytics/uptime_data.go
vaske: I think @furkansenharputlu was about to remove all `Codexxx` and make it as one property in struct but as map, this makes sense to me

main.go
```diff
@@ -345,7 +361,15 @@ func main() {
...
-
+	/*
+		if *demoUptime != "" && !SystemConfig.DontPurgeUptimeData{
+			log.Info("BUILDING UPTIME DEMO DATA AND EXITING...")
+			log.Warning("Starting from date: ", time.Now().AddDate(0, 0, -30))
+			demo.DemoInit(*demoUptime, *demoApiMode, *demoApiVersionMode)
+			demo.GenerateDemoUptimeData(time.Now().AddDate(0, 0, -30), 30, *demoUptime, UptimePump.WriteUptimeData)
+
+			return
+		}*/
```
vaske, line 372: I think we can remove this if it's not in use, at least you always have git history to revert it back :)

analytics/uptime_data.go
furkansenharputlu: Thanks @vaske! I meant exactly same.

tbuchaillot: @furkansenharputlu @vaske  those `code_xx` are **counters columns** in the SQL database, that's why they are like these. if we go with a map structure, we're going to miss that or Am I missing something?

main.go
tbuchaillot: @vaske good catch! I was trying something there and forgot to delete it 👍

pumps/sql_test.go
```diff
@@ -119,3 +120,161 @@ func TestSQLWriteDataSharded(t *testing.T) {
 	assert.Len(t, dbRecords, 2)
 
 }
+
+func TestSQLWriteUptimeData(t *testing.T) {
+	pmp := SQLPump{IsUptime: true}
+	cfg := make(map[string]interface{})
+	cfg["type"] = "sqlite"
+	cfg["connection_string"] = "pmp_test.db"
```
furkansenharputlu, line 128: maybe we can try in-mem to be consistent with other tests.

pumps/sql_test.go
furkansenharputlu: This code is repeating, we can make it modular and create a test func which accepts data and expected length etc.

```encoded, _ = msgpack.Marshal(analytics.UptimeReportData{OrgID: "1", URL: "url1", TimeStamp: nowPlus1})
	keys[0] = string(encoded)
	keys[1] = string(encoded)
	keys[2] = string(encoded)

	pmp.WriteUptimeData(keys)

	dbRecords = []analytics.UptimeReportAggregateSQL{}

	if err := pmp.db.Table(table).Find(&dbRecords).Error; err != nil {
		t.Fatal("Error getting analytics records from SQL")
	}
	assert.Len(t, dbRecords, 4)
```

analytics/uptime_data.go
furkansenharputlu: Can't we create a code struct and use that? We are using same structure for `SQLAnalyticsRecordAggregate` and `ProcessStatusCodes` is the same function redundantly. Also, `UptimeReportAggregateSQL` and `SQLAnalyticsRecordAggregate` look like same, why do we create a new struct like this?

analytics/uptime_data.go
```diff
@@ -1,20 +1,332 @@
...
+		dimensions = append(dimensions, Dimension{"errors", key, inc})
+	}
+
+	dimensions = append(dimensions, Dimension{"", "total", &u.Total})
+
+	return
+}
+
+func (a *UptimeReportAggregateSQL) TableName() string {
+	return "tyk_uptime_analytics"
```
furkansenharputlu, line 116: We are using this `tyk_uptime_analytics` throughout the PR, we can put it as a `const`

main.go
```diff
@@ -156,18 +156,33 @@ func initialisePumps() {
...
+
+	default:
+		UptimePump = &pumps.MongoPump{IsUptime: true}
+		UptimePump.Init(SystemConfig.UptimePumpConfig.MongoConf)
+	}
+	log.WithFields(logrus.Fields{
+		"prefix": mainPrefix,
+	}).Info("'dont_purge_uptime_data' set to false, attempting to start Uptime pump! ", UptimePump.GetName())
+
+	//UptimePump.Init(SystemConfig.UptimePumpConfig)
```
furkansenharputlu, line 179: leftover ?

pumps/sql_test.go
tbuchaillot: Done @furkansenharputlu

pumps/sql_test.go
tbuchaillot: Done

analytics/uptime_data.go
tbuchaillot: Done

main.go
tbuchaillot: Done

analytics/uptime_data.go
tbuchaillot: @furkansenharputlu I tried to embed  `SQLAnalyticsRecordAggregate` since it has similar behavior + fields but for some reason, gorm is not using the `gorm:"embedded"` tag and it always adds the struct name as a prefix. For example, for the Counter struct, it adds `counter_XXXX` in each column.

I think the best is to let it like this at this moment and not change everything (tests + dashboard) for this.

analytics/analytics.go
```diff
@@ -24,6 +24,8 @@ type Latency struct {
 	Upstream int64 `json:"upstream"`
 }
 
+const SQLTABLE = "tyk_analytics"
```
furkansenharputlu, line 27: In Golang, const naming convention is like: `SQLTable` and `UptimeSQLTable`. I will approve it anyway but wanted to mention if you'd like to change.

furkansenharputlu (approved):

analytics/analytics.go
tbuchaillot: Done!

PR #365, opened 2021-06-29
[TT-2783] Implement table sharding for SQLAggregatePump

furkansenharputlu:
This PR implements table sharding in `SQLAggregatePump`. The logic is a bit different when you compare with other pumps' table sharding. This PR handles everything within one for loop. Also, it doesn't make auto migration separately for the first and second days, instead, it make migration when it is necessary.

### DIFF
--- a/pumps/sql_aggregate.go
+++ b/pumps/sql_aggregate.go
@@ -97,24 +97,55 @@ func (c *SQLAggregatePump) Init(conf interface{}) error {
 func (c *SQLAggregatePump) WriteData(ctx context.Context, data []interface{}) error {
 	c.log.Debug("Attempting to write ", len(data), " records...")
 
-	analyticsPerOrg := analytics.AggregateData(data, c.SQLConf.TrackAllPaths, c.SQLConf.IgnoreTagPrefixList, c.SQLConf.StoreAnalyticsPerMinute)
+	var typedData []*analytics.AnalyticsRecord
 
-	for orgID, ag := range analyticsPerOrg {
-		// ag.DiscardAggregations([]string{"keyendpoints", "oauthendpoints", "apiendpoints"})
+	batch := 500
+	newBatch := false
+	for i, r := range data {
+		rec := r.(analytics.AnalyticsRecord)
+		typedData = append(typedData, &rec)
 
-		for _, d := range ag.Dimensions() {
-			rec := analytics.SQLAnalyticsRecordAggregate{
-				OrgID:          orgID,
-				TimeStamp:      ag.TimeStamp.Unix(),
-				Counter:        *d.Counter,
-				Dimension:      d.Name,
-				DimensionValue: d.Value,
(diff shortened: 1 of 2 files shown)


pumps/sql_aggregate_test.go
```diff
@@ -35,6 +37,59 @@ func TestSQLAggregateInit(t *testing.T) {
 
 }
 
+func TestSQLAggregateWriteData_Sharded(t *testing.T) {
+	pmp := SQLAggregatePump{}
+	cfg := make(map[string]interface{})
+	cfg["type"] = "sqlite"
+	cfg["connection_string"] = "pmp_test.db"
+	cfg["table_sharding"] = true
```
vaske, line 45: should we also add test with Postgres or we are fine with just sqlite?

furkansenharputlu: The existing tickets were just for sqlite so I went with it. Also, there is no postgres running in ci-tests.

pumps/sql_aggregate.go
```diff
@@ -95,31 +95,58 @@ func (c *SQLAggregatePump) Init(conf interface{}) error {
...
 			}
-			rec.ProcessStatusCodes()
+		} else {
+			i = dataLen // write all records at once for non-sharded case, stop for loop after 1 iteration
+		}
 
-			resp := c.db.WithContext(ctx).Create(rec)
-			if resp.Error != nil {
-				return resp.Error
+		analyticsPerOrg := analytics.AggregateData(data[startIndex:endIndex], c.SQLConf.TrackAllPaths, c.SQLConf.IgnoreTagPrefixList, c.SQLConf.StoreAnalyticsPerMinute)
```
buger, line 126: That's smart approach to deal with multi-hour records. 
But it took me some time to understand how it works.
Maybe makes sense to extend code comments a bit.

pumps/sql_aggregate_test.go
buger: In this case why to rely on database file at all. Maybe just keep connection string empty and use in-memory db?

furkansenharputlu: Not sure what you mean by in-memory db? Does our new engine have such a feature when it is empty? Also, shouldn't we test with at least one SQL database?

buger: Yes, if you leave it empty, SQLite switch to in-memory mode.

furkansenharputlu: Is in-memory a real SQL environment?

tbuchaillot (approved):
LGTM. As leo said, some comments in the tablesharding code would be 👌

pumps/sql_aggregate_test.go
furkansenharputlu: I removed test file and use in-memory now.

pumps/sql_aggregate.go
furkansenharputlu: added.

pumps/sql_aggregate_test.go
furkansenharputlu: done

PR #377, opened 2021-07-30
Splunk Pump - More analytic fields and tags ignore option

mbelchev:
## Description

With this pull request, we're extending the Tyk Splunk Pump by adding more fields that can be extracted from each analytics record.

One of the newly added fields is the `tags`. Alongside adding it as an option, we're introducing a new config option in order to provide a way for filtering the tags before sending them to Splunk.

The documentation has also been updated in this pull request in order to match the changes in it.

## Related Issue

[#13011](https://support.tyk.io/hc/en-gb/requests/13011)

## Motivation and Context

Making the Tyk Splunk Pump even more powerful - adding new fields and new config options!

## How This Has Been Tested

The suggested feature has been tested in a real and active environment in our company. For more details, we can discuss in the related ticket.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/pumps/splunk.go
+++ b/pumps/splunk.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"time"
+	"strings"
 
 	"github.com/mitchellh/mapstructure"
 
@@ -109,6 +110,7 @@ type SplunkPumpConfig struct {
 	ObfuscateAPIKeys       bool     `mapstructure:"obfuscate_api_keys"`
 	ObfuscateAPIKeysLength int      `mapstructure:"obfuscate_api_keys_length"`
 	Fields                 []string `mapstructure:"fields"`
+	IgnoreTagPrefixList    []string `mapstructure:"ignore_tag_prefix_list"`
 }
 
 // New initializes a new pump.
@@ -184,7 +186,11 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
 			"raw_response":   decoded.RawResponse,
 			"ip_address":     decoded.IPAddress,
 			"geo":            decoded.Geo,
+			"network":        decoded.Network,
+			"latency":        decoded.Latency,
+			"tags":           decoded.Tags,
 			"alias":          decoded.Alias,
+			"track_path":     decoded.TrackPath,
 		}
 
 		// Define an empty event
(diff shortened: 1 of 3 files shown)


n2linux:
@mbelchev Hi Mariyan - thanks so much for doing this work and submitting the PR. Someone from the team review it and we can go from there.

tbuchaillot (approved):
Looks good @mbelchev! Thanks for your contribution!

gregdelhon:
@mbelchev Thanks for this, great contrib. How did you find it? Anything we can do to make this better? 
Can you share more on how you approached testing, I can see it's already in use in an environment. It will help fast-track our QA & Release process!

pumps/splunk.go
```diff
@@ -199,6 +205,27 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
...
+				// Check if the current analytics field is "tags" and see if some tags are explicitly excluded
+				if (field == "tags" && len(p.config.IgnoreTagPrefixList) > 0) {
+					filteredTags := mapping["tags"].([]string)
+
+					// Loop all explicitly ignored tags
+					for _, excludeTag := range p.config.IgnoreTagPrefixList {
+						// Loop the current analytics item tags
+						for key, currentTag := range filteredTags {
+							// If the current tag's value includes an ignored word, remove it from the list
+							if strings.Contains(currentTag, excludeTag) {
```
tbuchaillot, line 217: @mbelchev  It should use `strings.HasPrefix` instead of contains.

pumps/splunk.go
```diff
@@ -199,6 +205,27 @@ func (p *SplunkPump) WriteData(ctx context.Context, data []interface{}) error {
 					continue
 				}
 
+				// Check if the current analytics field is "tags" and see if some tags are explicitly excluded
+				if (field == "tags" && len(p.config.IgnoreTagPrefixList) > 0) {
+					filteredTags := mapping["tags"].([]string)
+
+					// Loop all explicitly ignored tags
+					for _, excludeTag := range p.config.IgnoreTagPrefixList {
```
tbuchaillot, line 213: @mbelchev this could be in a separated method for `SplunkPump` and have its own unit test. What do you think?

something like 

```go
func (p *SplunkPump) FilterTags(filteredTags []string) {
	// Loop all explicitly ignored tags
	for _, excludeTag := range p.config.IgnoreTagPrefixList {
		// Loop the current analytics item tags
		for key, currentTag := range filteredTags {
			// If the current tag's value includes an ignored word, remove it from the list
			if strings.HasPrefix(currentTag, excludeTag) {
				copy(filteredTags[key:], filteredTags[key+1:])
				filteredTags[len(filteredTags) - 1] = ""
				filteredTags = filteredTags[:len(filteredTags)-1]
			}
		}
	}
}
```

tbuchaillot (changes requested):

tbuchaillot:
All the info moved to https://tyktech.atlassian.net/browse/TT-3211. I'll close this PR and raise a new one when this is implemented.

pumps/splunk.go
mbelchev: @tbuchaillot, yep, agree. It's now adjusted, thanks!

pumps/splunk.go
mbelchev: Good point, it's now moved in a separate function. Thank you

PR #387, opened 2021-08-30
TT-3243 SQL Aggregate and Uptime pump write optimization

tbuchaillot:
## Description

Changed the way we write SQL Aggregations and Uptime SQL Aggregations. Now, we write them in batch (all the records together in 1 shot) and we use the special `excluded` table for  `OnConflict statements`.  How does that work?  The `SET` clause in `ON CONFLICT DO UPDATE` has access to the existing row using the table name (`tyk_aggregated` in our case) but also has access to the proposed insertion in the `excluded` table.  
https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT
https://www.sqlite.org/lang_UPSERT.html

Added `batch_size` configuration option for all SQL pumps, by default it writes 1000 records max.

**Doing that we went from 6-7 seconds write to 1 sec / <1sec writes.**

## Related Issue

https://tyktech.atlassian.net/browse/TT-3243
## How This Has Been Tested

Tested with the current unit tests to  avoid breaking anything  + using `demo flag --demo="60e321d89858dc6c61e17b4a" --demo-api="ef33ff3dbc2543a547204fce5aa08a72"`

With optimization:
image

Without optimization:
image

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


sredxny (approved):
lgtm! lets merge it

matiasinsaurralde (approved):

PR #388, opened 2021-08-31
TT-2926 Feature/Graceful shutdown

tbuchaillot:
## Description

Adding a mechanism to gracefully shutdown tyk-pump  and adding a `Shutdown` func for all the pumps. 
The only two pumps that do something in that func is `ES`and `dogstatsd` pump, which will flush all the records if bulk config is true.

This feature also implies that on shutdown, tyk-pump is going to wait until the current `purge_loop` cycle finish before ending the process. That means that we will not lose the current processing data on process interrupt/shutdown.

Tyk-pump is going to shut down gracefully on: `os.Interrupt, syscall.SIGINT, syscall.SIGTERM`.
 

## Related Issue

https://tyktech.atlassian.net/browse/TT-2926
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


sredxny (approved):
LGTM! 🚀

letzya:
Looking good. Maybe a few tests?

buger:
I do not  think it should do anything on SIGKILL signal. 
The purpose of this signal is to force process shutdown immediately.

@gwotg any thoughts on it from DevOps point of view?

gwotg:
Think you're right @buger - unless on the parent process, trying to catch or handle sigkill shouldn't work - if we can, this is to be considered an OS bug to be reported (lol). Even so, there isn't much point - if we reached the kill point, let it go - we'll inform the customer on the graceful termination process in the docs (and maybe explicitly inform that sigkill == undefined behavior).

Ref: 
https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html
https://en.wikipedia.org/wiki/Signal_(IPC)#Handling_signals

tbuchaillot:
good catch @buger @gwotg! thanks!

tbuchaillot:
@letzya added a test for this scenario 👍

matiasinsaurralde (approved):
LGTM

buger (approved):
👍🏻

sredxny (approved):
LGTM 🚀

PR #402, opened 2021-11-10
Decoupling aggregated writing

tbuchaillot:
## Description

Decoupling aggregated writing from `WriteData` method in SQL and Mongo aggregate pumps in order to make it possible to embed them in MDCB.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/mongo_aggregate.go
+++ b/pumps/mongo_aggregate.go
@@ -254,7 +254,7 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{}) 
 				return err
 			}
 
-			m.log.Debug("Wrote aggregated data for ", len(data), " records")
+			m.log.Debug("Processed aggregated data for ", orgID)
 		}
 	}
 	m.log.Info("Purged ", len(data), " records...")
--- a/pumps/mongo_aggregate.go
+++ b/pumps/mongo_aggregate.go
@@ -249,78 +249,12 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{}) 
 
 		// put aggregated data into MongoDB
 		for orgID, filteredData := range analyticsPerOrg {
-			collectionName, collErr := m.GetCollectionName(orgID)
-			if collErr != nil {
-				m.log.Info("No OrgID for AnalyticsRecord, skipping")
-				continue
-			}
-
-			thisSession := m.dbSession.Copy()
-			defer thisSession.Close()
-
-			analyticsCollection := thisSession.DB("").C(collectionName)
-			indexCreateErr := m.ensureIndexes(analyticsCollection)
-
-			if indexCreateErr != nil {
-				m.log.Error(indexCreateErr)
-			}
-
-			query := bson.M{
-				"orgid":     filteredData.OrgID,
-				"timestamp": filteredData.TimeStamp,
-			}
-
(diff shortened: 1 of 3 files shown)


pumps/mongo_aggregate.go
```diff
@@ -249,78 +249,12 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{})
...
-			if m.dbConf.UseMixedCollection {
-				thisData := analytics.AnalyticsRecordAggregate{}
-				err := analyticsCollection.Find(query).One(&thisData)
-				if err != nil {
-					m.log.WithField("query", query).Error("Couldn't find query doc:", err)
-				} else {
-					m.doMixedWrite(thisData, query)
-				}
-			}
+			m.log.Debug("Wrote aggregated data for ", len(data), " records")
```
sredxny, line 257: shouldn't it be `len(filteredData)` ?

pumps/mongo_aggregate.go
```diff
@@ -373,6 +307,78 @@ func (m *MongoAggregatePump) HandleWriteErr(err error) error {
 	return err
 }
 
+func (m *MongoAggregatePump) DoAggregatedWriting(ctx context.Context, orgID string, filteredData analytics.AnalyticsRecordAggregate) error {
+	collectionName, collErr := m.GetCollectionName(orgID)
+	if collErr != nil {
+		m.log.Info("No OrgID for AnalyticsRecord, skipping")
+		return nil
```
sredxny, line 314: if this returns nil, then the caller function will say that the records were added, which would be weird as they were not. Is an issue to return an error? or some boolean like: `didInsert`

pumps/mongo_aggregate.go
```diff
@@ -373,6 +307,78 @@ func (m *MongoAggregatePump) HandleWriteErr(err error) error {
...
+	}
+
+	thisSession := m.dbSession.Copy()
+	defer thisSession.Close()
+
+	analyticsCollection := thisSession.DB("").C(collectionName)
+	indexCreateErr := m.ensureIndexes(analyticsCollection)
+
+	if indexCreateErr != nil {
+		m.log.Error(indexCreateErr)
```
sredxny, line 324: should we return err?

sredxny (changes requested):

pumps/mongo_aggregate.go
tbuchaillot: @sredxny before it didn't cut the writing execution if the was an error creating indexes. I don't want to change that logic, IMO we should keep trying to write the records despite the index

pumps/mongo_aggregate.go
tbuchaillot: @sredxny nop, since filteredData is the already aggregated records. But I agree that `len(data)` is not correct. 

What do you think if we log `m.log.Debug("Wrote aggregated data for " + orgID)` ?

sredxny: @tbuchaillot that would be better

pumps/mongo_aggregate.go
sredxny: ok, got it

sredxny (approved):
lgtm

PR #409, opened 2021-12-28
TD-664/Aws-cf-templates

ermirizio:
M4 template sync for TD-664 branch AWS CF templates


ermirizio:
Not sure about the releases where the original commit needs to merged, if 1.5 makes sense or not, what else?

alephnull (approved):

alephnull:
> Not sure about the releases where the original commit needs to merged, if 1.5 makes sense or not, what else?

I think just 1.5 and master is good enough for now.

PR #415, opened 2022-01-20
Sync master with m4 release engineering templates

ermirizio:
Sync m4 templates


.goreleaser.yml
```diff
@@ -12,11 +12,10 @@
 # - macOS (only 10.15 is supported)
 # - amd64
 
-
 builds:
   - id: std
     ldflags:
-      - -X main.VERSION={{.Version}} -X main.commit={{.FullCommit}} -X main.buildDate={{.Date}} -X main.builtBy=goreleaser
+      - -X github.com/TykTechnologies/tyk-pump/pumps.VERSION={{.Version}} -X github.com/TykTechnologies/tyk-pump/pumps.commit={{.FullCommit}} -X github.com/TykTechnologies/tyk-pump/pumps.buildDate={{.Date}} -X github.com/TykTechnologies/tyk-pump/pumps.builtBy=goreleaser
```
excieve, line 18: For this repo, it actually appears to be on the `main` package :/
https://github.com/TykTechnologies/tyk-pump/blob/master/version.go

Also `commit` is actually `Commit` (similar in other repos it seems).

excieve (changes requested):
Thanks @ermirizio! Looks like `VERSION` and other build-time variables are still in the `main` package here, so part of this change won't work.

PR #418, opened 2022-01-24
add missing version.go

ermirizio:
Add mssing version.go in order to normalice design as done in tyk,dashboard, etc


tbuchaillot (approved):

tbuchaillot (changes requested):
@ermirizio you need to update the `VERSION` reference from main.go, since it was moved to another pkg

tbuchaillot (approved):
LGTM

PR #432, opened 2022-03-04
Update sync automation

asutosh:
## Description
Testing sync automation workflow trigger issue on push.


.github/workflows/sync-automation.yml
```diff
@@ -36,6 +38,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{matrix.branch}}
+          token: ${{ secrets.ORG_GH_TOKEN }}
```
alephnull, line 41: I think the token is required on L52 below, not here.

alephnull (changes requested):

.github/workflows/sync-automation.yml
```diff
@@ -38,9 +38,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{matrix.branch}}
-          token: ${{ secrets.ORG_GH_TOKEN }}
 
       - name: sync ${{matrix.branch}} from master
+        env:
+          GITHUB_TOKEN: ${{ secrets.ORG_GH_TOKEN }}
```
asutosh, line 44: @alephnull Have added by setting the GITHUB_TOKEN env variable in the push step - not completely sure if this would work? Other option is to set the url with the token included, but is it okay since that would be set in the .git/config file in filesystem?

alephnull: Yeah, this change by itself is unlikely to help. This token only works for http so git will have to be setup to use it.

It appears that using deploy keys does trigger the Release action so you could try implementing that if you stuck for options. However, this is a last resort as there we will have to manage deploy keys for each repo individually afaik.

Sadly, to understand the behaviour, I see no other option than to try things.

alephnull: From https://github.community/t/push-from-action-even-with-pat-does-not-trigger-action/17622/6 it appears that this change needs `persist-credentials: false` for checkout@v2.

asutosh: In that case, if we set the GITHUB_TOKEN in the checkout stage to `secrets.ORG_GH_TOKEN`, that will be persisted in this stage then?

alephnull: There's only one way to be sure.

asutosh: I'll try with a test repo and test action and confirm.

alephnull (approved):

PR #405, opened 2021-12-07
Feature/mongo pump indexes

tbuchaillot:
## Description

Avoid Mongo Pumps index recreation if the collection exists. This will allow to delete / change the mongo pump indexes manually without tyk-pump re-creating them each time.
This applies for mongo pump, mongo aggregate pump (with mixed collections and without) and mongo selective pump .

## Motivation and Context

https://tyktech.atlassian.net/browse/TT-3809

## How This Has Been Tested

Configure 

1 - Configure mongo pumps, for example:
```
 "pumps": {
         "mongo": {
          "type": "mongo",
          "meta": {
            "collection_name": "tyk_analytics",
            "mongo_url": "mongodb://localhost/tyk_analytics"
          }
        },
        "mongo-pump-aggregate": {
          "type": "mongo-pump-aggregate",
          "meta": {
            "mongo_url": "mongodb://localhost/tyk_analytics",
            "use_mixed_collection": true
          }
        }
    },
```
2 - Write some demo data ( run pump with  --demo="60e321d89858dc6c61e17b4a" --demo-api="ef33ff3dbc2543a547204fce5aa08a72"       )
3 - Check if the mongo collections and indexes are created.
4 - Stop pump
5 - Delete the indexes manually in mongo
6 - Run pump with a demo flag again, **it shouldn't recreate the indexes**.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If pulling from your own
      fork, don't request your `master`!
- [ ] Make sure you are making a pull request against the **`master` branch** (left side). Also, you should start
      *your branch* off *our latest `master`*.
- [ ] My change requires a change to the documentation.
  - [ ] If you've changed APIs, describe what needs to be updated in the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] Check your code additions will not fail linting checks:
  - [ ] `go fmt -s`
  - [ ] `go vet`

PR #442, opened 2022-05-12
TT-5302 Mongo pump indexes on DocumentDB

tbuchaillot:
## Description

This PR adds a configuration option for Mongo pumps to avoid the index creation directly `omit_index_creation`. 

It also adds an exception for the change added in #405, to avoid the index no-creation if the collection exists in DocumentDB. I made that second change since the `CollectionNames` function from mgo lib, always returns an empty string slice in DocDB and also leaks cursors.

So basically, from now on, the index behavior is the following:
- If `omit_index_creation` is set to true, it won't create indexes for mongo pumps.
- If `omit_index_creation` is set to false (default) and we are using `DocumentDB`, it's always going to create the indexes.
- If `omit_index_creation` is set to false (default) and we are using `MongoDB`, we have 2 behaviors:
  -  if the collection exists, we are not going to create the indexes again.
  - if the collection doesn't exist, we are going to create the indexes.

## Related Issue

https://tyktech.atlassian.net/browse/TT-5302

## Motivation and Context

https://tyktech.atlassian.net/browse/TT-5302

## How This Has Been Tested

Added unit tests mocking different behaviors.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


pumps/mongo.go
```diff
@@ -392,12 +394,21 @@ func (m *MongoPump) collectionExists(name string) (bool, error) {
 }
 
 func (m *MongoPump) ensureIndexes() error {
-	exists, errExists := m.collectionExists(m.dbConf.CollectionName)
-	if errExists == nil && exists {
-		m.log.Debug("Collection ", m.dbConf.CollectionName, " exists, omitting index creation")
+	if m.dbConf.OmitIndexCreation {
+		fmt.Println("m.dbConf.OmitIndexCreation", m.dbConf.OmitIndexCreation)
```
sredxny, line 398: remove this line or convert to log.Debug

pumps/mongo_test.go
```diff
@@ -32,6 +33,103 @@ func TestMongoPump_capCollection_Enabled(t *testing.T) {
...
+			}
+
+			mPump.WriteData(context.Background(), records)
+			indexes, errIndexes := c.GetIndexes()
+			if errIndexes != nil {
+				t.Error("error getting indexes:", errIndexes)
+			}
+
+			if len(indexes) != tc.Indexes {
+				t.Error("index differential:", indexes)
```
sredxny, line 127: what does it means `index diferential`? can we change it to something more explanatory?

sredxny (changes requested):
added a few comments, everything else looks good!

pumps/mongo_test.go
tbuchaillot: Changed the message @sredxny 👍

sredxny (approved):
lgtm

PR #445, opened 2022-06-06
TT-5139 Refactor Prometheus/WriteData method to improve performance

sredxny:
## Description
Refactored the method to avoid to call the method `inc` on each loop operation, and call it only once at the end of the function as internally the library is using atomic variables, so, is better to avoid so repetitives assigning calls.

## Related Issue
https://tyktech.atlassian.net/browse/TT-5139

## Motivation and Context
Give solution to client

## How This Has Been Tested
- Ran app

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/prometheus.go
+++ b/pumps/prometheus.go
@@ -3,10 +3,11 @@ package pumps
 import (
 	"context"
 	"errors"
+	"fmt"
+	"github.com/TykTechnologies/tyk-pump/analytics"
 	"net/http"
 	"strconv"
-
-	"github.com/TykTechnologies/tyk-pump/analytics"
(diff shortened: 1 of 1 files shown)


tbuchaillot:
@sredxny I think it's going to be better if we make something like `map[string]struct{}` instead of having different maps ( this will collide with the custom Prometheus metrics PR). 
We can define the struct with a metric type/fields and a counter. What do you think?

tbuchaillot:
closed for https://github.com/TykTechnologies/tyk-pump/pull/452

PR #453, opened 2022-06-20
TT-5776 Racy filterData func

tbuchaillot:
## Description

This PR changes the way we duplicate the slice while filtering. The previous way was more mem efficient but it was racy when using multiple pumps.

This PR also adds more tests scenarios for filtering data.

## Related Issue

https://tyktech.atlassian.net/browse/TT-5776
## Motivation and Context

https://tyktech.atlassian.net/browse/TT-5776
## How This Has Been Tested

Added unit test. 
Ran test with -race

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


main.go
```diff
@@ -299,10 +299,13 @@ func filterData(pump pumps.Pump, keys []interface{}) []interface{} {
 	if !filters.HasFilter() && !pump.GetOmitDetailedRecording() && !shouldTrim {
 		return keys
 	}
-	filteredKeys := keys[:]
+
+	filteredKeys := make([]interface{}, len(keys))
```
tbuchaillot, line 303: this is the brain of the fix. In concurrent scenarios - and with filters, when modifying `filteredKeys` we are modifying the `keys` slice.

With this solution, we are copying all the keys in this new slice (only if it has filters or omit_detailed_recording) in this way we don't override the original slice.

buger: I wonder if for performance reasons we should do this copy ONLY if we have at least one filter?

tbuchaillot: @buger that's already happening.

we return the same slice without copying it if it doesn't have filters / or need some general sanitization https://github.com/TykTechnologies/tyk-pump/blob/027a664557c7f095fef3383614bc8b60619fa520/main.go#L299

buger (approved):
Makes sense!

Love the tests!

PR #463, opened 2022-07-28
ignore graph analytics for mongo pump(s)

pvormste:
This PR adds the behavior of ignoring graph analytics records. This means that records that carry the tag `tyk-graph-analytics` won't be stored in the `tyk-analytics` and `tyk_analytics_aggregate` collection (if using default names).

It's a non-breaking change and the new pump to handle graph analytics will be following shortly.

*Why this change?*
Graph records are different from classic REST records. Instead of endpoints graphs are dealing with operations and operation paths. 
HTTP status codes are less relevant for errors, instead, the errors are part of the body response.
All that means that a different struct is necessary to store this data - so we don't need to store graph data in the current collections.

[TT-6121](https://tyktech.atlassian.net/browse/TT-6121)

## Types of changes

### DIFF
--- a/analytics/aggregate.go
+++ b/analytics/aggregate.go
@@ -539,7 +539,7 @@ func replaceUnsupportedChars(path string) string {
 }
 
 // AggregateData calculates aggregated data, returns map orgID => aggregated analytics data
-func AggregateData(data []interface{}, trackAllPaths bool, ignoreTagPrefixList []string, storeAnalyticPerMinute bool) map[string]AnalyticsRecordAggregate {
+func AggregateData(data []interface{}, trackAllPaths bool, ignoreTagPrefixList []string, storeAnalyticPerMinute, ignoreGraphData bool) map[string]AnalyticsRecordAggregate {
 	analyticsPerOrg := make(map[string]AnalyticsRecordAggregate)
 	for _, v := range data {
 		thisV := v.(AnalyticsRecord)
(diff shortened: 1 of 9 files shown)


analytics/aggregate.go
```diff
@@ -549,6 +549,11 @@ func AggregateData(data []interface{}, trackAllPaths bool, ignoreTagPrefixList [
 			continue
 		}
 
+		// We don't want to aggregate Graph Data with REST data - there is a different type for that.
+		if thisV.IsGraphRecord() {
```
tbuchaillot, line 553: @pvormste I think the best approach for aggregated records is to start receiving a bool flag and `if ignoreGraphQL && thisV.IsGraphRecord()` here. 
Why? Because there's another pump - `sql aggregate` that uses this function and it's going to generate different analytics between "normal" SQL and aggregated SQL.

kofoworola:
@tbuchaillot the demo data generator will be modified in a follow up PR that implements the mongo-graph pump

analytics/aggregate.go
kofoworola: Yeah so for now, i changed it so only data trying to be aggregated by the mongo aggregate pump is filtered.

buger:
@kofoworola why not to use new "schema" field to check if it is a graph analytics?

tbuchaillot (approved):
LGTM

kofoworola:
> @kofoworola why not to use new "schema" field to check if it is a graph analytics?

Yeah that's a better idea, so leaving this PR unmerged for now.

PR #474, opened 2022-08-17
[TT-6012] add ApiSchema to analytics record

kofoworola:
## Description
This PR adds an ApiSchema field to the `analytics.AnalyticsRecord` as well as the proto equivalent of the analytics record representation

## Related Issue

## Motivation and Context
We intend to add analytics support for graphql queries for information like fields/types/variables etc.. and for those to be extracted properly we need to add some extra data, more specifically, the schema of the request being made, which is fetched from the api spec in the gateway

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


buger:
I do not like Map used here, because it will be impossible to implement it properly in SQL. 
If we need fields like Type I woud rather add this fields directly to the struct.

kofoworola:
> I do not like Map used here, because it will be impossible to implement it properly in SQL. If we need fields like Type I woud rather add this fields directly to the struct.

Thanks for bringing this up, changed to a more type safe field.

buraksezer (approved):

buger (approved):

PR #475, opened 2022-08-18
TT-6343 Fixing mem address pointer in prom custom metrics

tbuchaillot:
## Description

Basically, the for loop always took the last memory address of the custom metrics config. That makes the custom metrics always take the last custom metric value. 

Needs to add unit tests before merging.

## Related Issue

https://tyktech.atlassian.net/browse/TT-6343
## How This Has Been Tested

Added unit tests - tested manually

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/prometheus.go
+++ b/pumps/prometheus.go
@@ -169,10 +169,8 @@ func (p *PrometheusPump) Init(conf interface{}) error {
 	}
 
 	//then we check the custom ones
-	errInitCustom := p.InitCustomMetrics()
-	if errInitCustom != nil {
-		p.log.Error("error initializing custom metrics:", errInitCustom)
-	}
+	p.InitCustomMetrics()
+
 	p.log.Info("Starting prometheus listener on:", p.conf.Addr)
 
 	http.Handle(p.conf.Path, promhttp.Handler())
@@ -186,7 +184,7 @@ func (p *PrometheusPump) Init(conf interface{}) error {
 }
 
 //InitCustomMetrics initialise custom prometheus metrics based on p.conf.CustomMetrics and add them into p.allMetrics
-func (p *PrometheusPump) InitCustomMetrics() error {
+func (p *PrometheusPump) InitCustomMetrics() {
 	if len(p.conf.CustomMetrics) > 0 {
 		customMetrics := []*PrometheusMetric{}
 		for i := range p.conf.CustomMetrics {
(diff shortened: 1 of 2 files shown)


pumps/prometheus.go
```diff
@@ -169,14 +169,32 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
 	if len(p.conf.CustomMetrics) > 0 {
 		customMetrics := []*PrometheusMetric{}
 		for i := range p.conf.CustomMetrics {
 			newMetric := &p.conf.CustomMetrics[i]
 			newMetric.aggregatedObservations = p.conf.AggregateObservations
 			errInit := newMetric.InitVec()
 			if errInit != nil {
-				p.log.Error(errInit)
+				return errInit
 			} else {
```
mativm02, line 198: A very little observation: this else statement is useless, as you're returning the error before if it exists

mativm02 (approved):
LGTM

pumps/prometheus.go
tbuchaillot: solved!

pumps/prometheus.go
```diff
@@ -197,6 +185,27 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
+//InitCustomMetrics initialise custom prometheus metrics based on p.conf.CustomMetrics and add them into p.allMetrics
+func (p *PrometheusPump) InitCustomMetrics() error {
+	if len(p.conf.CustomMetrics) > 0 {
+		customMetrics := []*PrometheusMetric{}
+		for i := range p.conf.CustomMetrics {
+			newMetric := &p.conf.CustomMetrics[i]
+			newMetric.aggregatedObservations = p.conf.AggregateObservations
+			errInit := newMetric.InitVec()
+			if errInit != nil {
+				return errInit
```
sredxny, line 197: @tbuchaillot if we have an error is ok that we just return? and do not continue in the next iteration?

tbuchaillot: @sredxny do you suggest that we just log the name of the invalid one and add the others?

tbuchaillot: @sredxny changed to work in a non-blocking when missconfigured. good suggestion

sredxny (approved):
lgtm

PR #481, opened 2022-09-01
TT-3514 - Adding CosmosDB support

mativm02:
## Description

Adding the option of connecting to CosmosDB
Avoiding the `expireAt` index, because CosmosDB does not support it.

I realized that there are other options to fix this:
1. We can also add the `_ts` index only for CosmosDB, instead of `expireAt`. But, if we are just setting it to 0 (non-expire), I think it is unnecessary because it is 0 by default. Also, we should add the _ts index to the AnalyticsRecords struct, and only fill this field when the database is CosmosDB. 
2. We can just delete that part of the code if the idea is to only set it to 0 (non-expire). By default it will be set to 0, it doesn't matter if we hard-coded it or not. 

At the moment, this solution is working and setting the expireAt index to non-expire.

## Related Issue

https://tyktech.atlassian.net/browse/TT-3514
## Motivation and Context

It allows users to use Mongo's CosmosDB server.
## How This Has Been Tested

Unit tests were added.
I have also tested using Tyk's CosmosDB ([this is the used config](https://tyktech.atlassian.net/wiki/spaces/EN/pages/1329135632/External+DB+s)). I have created a testing API and analytics are being created, when a request succeeds or fails.
I did the same with a local MongoDB to check if everything was working like it was working before this change, and I didn't have any problem.

I ran units tests already created as well, to check functionality with all MongoDB's supported servers (including AWS DocumentDB)
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/mongo.go
```diff
@@ -164,9 +165,14 @@ func mongoType(session *mgo.Session) MongoType {
 
 	if result.Code == 303 {
 		return AWSDocumentDB
-	} else {
-		return StandardMongo
 	}
+
+	if result.Code == 115 {
+		return CosmosDB
+	}
```
sredxny, line 172: @mativm02 could you explain why 115 is CosmoDB?

mativm02: > @mativm02 could you explain why 115 is CosmoDB?

I saw that we were using the result code to check whether the connection is over DocumentDB or not. So I check what was the returned code when connecting to CosmosDB and it is 115. 
I tried to find a way to get some explicit information about the session (if it's connected to a simple Mongo, DocumentDB, or CosmosDB) but didn't find anything.

The method session.LiveServers() returns a part of the connection URL, in this case: `4-0-qa.mongo.cosmos.azure.com:10255`, I think we could check what's the DB server with it,  but I am not sure if this will work for all connections.

Do you know about another better option?

sredxny, line 172: @mativm02 lets convert it into a switch statement, and also what are your toughts about:

```
const(
     AWSDBError = 303
     CosmosDBError = 115
)

switch result.code {
   case AWSError:
      //blahblahblah
      case CosmosDBError:
      // blahblahblah
      Default:
      // StandardMongo
}
```

sredxny (changes requested):

sredxny (approved):
lgtm

pumps/mongo.go
```diff
@@ -68,7 +74,7 @@ type BaseMongoConf struct {
 	// Path to the PEM file which contains both client certificate and private key. This is
 	// required for Mutual TLS.
 	MongoSSLPEMKeyfile string `json:"mongo_ssl_pem_keyfile" mapstructure:"mongo_ssl_pem_keyfile"`
-	// Specifies the mongo DB Type. If it's 0, it means that you are using standard mongo db, but if it's 1 it means you are using AWS Document DB.
+	// Specifies the mongo DB Type. If it's 0, it means that you are using standard mongo db, if it's 1 it means you are using AWS Document DB, if it's 2, it means you are using CosmosDB.
```
tbuchaillot, line 77: @mativm02 can we add this info to the README and docs?

pumps/mongo_aggregate.go
```diff
@@ -221,15 +221,18 @@ func (m *MongoAggregatePump) ensureIndexes(c *mgo.Collection) error {
 	}
 
 	var err error
-	ttlIndex := mgo.Index{
-		Key:         []string{"expireAt"},
-		ExpireAfter: 0,
-		Background:  m.dbConf.MongoDBType == StandardMongo,
-	}
+	// CosmosDB does not support "expireAt" option
```
tbuchaillot, line 224: @mativm02 can we add that expireAt index is not supported in CosmoDB and it will be skipped in the README / docs?

tbuchaillot (changes requested):

tbuchaillot (approved):

tbuchaillot (changes requested):
@mativm02 please have a loot at the failing tests.

tbuchaillot (approved):

PR #486, opened 2022-09-08
[TT-506] - Self-Healing when hitting 16mb + making aggregation per time setting configurable 

mativm02:
## Description

When a document size hits 16Mb, we will create a new document with the current timestamp and reduce the aggregation per time configuration by half. 
Also, users can set an initial aggregation per time, using `analytics_stored_per_minute` in the .conf file, for example:
image
**Default value is set to 60.
The scope of the Jira Ticket is to fix this error when using Mongo Aggregate (it also supports Hybrid). SQL Aggregate should work as always and the idea is to make this change on it in another PR**

**Use cases:**
- If the user sets `analytics_stored_per_minute` = 40, Tyk-Pump will create an initial document on the first request, and will create another document when a new request is received and the current timestamp > last document timestamp + 40 minutes.
- If the current document hits 16Mb before 40 minutes have passed, a new document will be created and now, `analytics_stored_per_minute` will equal 20.
- If the current document hits 16Mb and `analytics_stored_per_minute` = 1, we can't reduce this setting, so the user will receive an error saying that it was unable to send the new data to the current document.
- If `analytics_stored_per_minute` is an odd number, we will round down (i.e: 3 will be converted to 1, not 1.5 or 2)
- Every time a document hits 16Mb and we heal it, we will fire an error log telling that there was a problem updating a document, and a warning log saying that we will try to fix it automatically. 

**Different Behaviors**
- If 2 or more pumps are writing data into the same collection, and the collection hits 16Mb, all of them are changing their aggregation time when they try to write data. This is happening because, when they try to write data, they will receive an error. Then, they will check if the error is related to the size of the document. Finally, each pump will activate the self-healing mechanism.
- SQL aggregate is working as always.
- It is currently working with 3 types of Mongo DBs: Standard Mongo, CosmosDB, DocumentDB. 

## Related Issue

https://tyktech.atlassian.net/browse/TT-506
## Motivation and Context

It allows users to set an aggregation per time. It also avoids errors when hitting max document size.
## How This Has Been Tested

Unit tests were created + manual testing using --demo flag to send multiple fake requests to APIs and send data to Redis to simulate the process.
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/hybrid.go
```diff
@@ -43,10 +43,10 @@ var (
 
 // HybridPump allows to send analytics to MDCB over RPC
 type HybridPump struct {
-	aggregated             bool
-	trackAllPaths          bool
-	storeAnalyticPerMinute bool
-	ignoreTagPrefixList    []string
+	aggregated               bool
+	trackAllPaths            bool
+	analyticsStoredPerMinute int
```
tbuchaillot, line 48: We cannot change configuration fields type - it would break backwards compatibility

pumps/mongo_aggregate.go
```diff
@@ -49,7 +49,8 @@ type MongoAggregateConf struct {
 	// Defaults to 1000.
 	ThresholdLenTagList int `json:"threshold_len_tag_list" mapstructure:"threshold_len_tag_list"`
 	// Determines if the aggregations should be made per minute instead of per hour.
-	StoreAnalyticsPerMinute bool `json:"store_analytics_per_minute" mapstructure:"store_analytics_per_minute"`
+	// Determines the amount of time the aggregations should be made. Max (and default) = 60 - Min = 1.
+	AnalyticsStoredPerMinute int `json:"analytics_stored_per_minute" mapstructure:"analytics_stored_per_minute"`
```
tbuchaillot, line 53: We cannot change configuration fields type - it would break backwards compatibility

tbuchaillot (changes requested):
@mativm02 some initial comments.

pumps/mongo_aggregate.go
mativm02: Now both configuration fields are compatible. If StoreAnalyticsPerMinute is not set (or it's false), we will look for the value of AggregationTime. If it does not exist either, set it to 60 by default.

pumps/mongo_aggregate.go
```diff
@@ -170,12 +177,32 @@ func (m *MongoAggregatePump) Init(config interface{}) error {
 	if m.dbConf.ThresholdLenTagList == 0 {
 		m.dbConf.ThresholdLenTagList = THRESHOLD_LEN_TAG_LIST
 	}
+	// if StoreAnalyticsPerMinute is set to true, the aggregation time will be set to 1.
+	// if not, the aggregation time will be set to the value of the field AggregationTime.
+	// if there is no value for AggregationTime, it will be set to 60.
+	if m.dbConf.StoreAnalyticsPerMinute {
+		m.log.Info("StoreAnalyticsPerMinute is set to true. Setting AggregationTime to 1.")
```
sredxny, line 184: lets be more informative here. Lets say something like: `Pump will aggregate data every minute.` or similar.

pumps/mongo_aggregate.go
```diff
@@ -48,8 +48,15 @@ type MongoAggregateConf struct {
 	// it will print an alert.
 	// Defaults to 1000.
 	ThresholdLenTagList int `json:"threshold_len_tag_list" mapstructure:"threshold_len_tag_list"`
-	// Determines if the aggregations should be made per minute instead of per hour.
+	// Determines if the aggregations should be made per minute (true) or per hour (false).
 	StoreAnalyticsPerMinute bool `json:"store_analytics_per_minute" mapstructure:"store_analytics_per_minute"`
+	// Determines the amount of time the aggregations should be made. Max (and default) = 60 - Min = 1.
```
sredxny, line 53: Let's rewrite this for something like: `Determines the amount of time the aggregations should be made (in minutes). It defaults to the max value is 60 and the minimum is 1.`

pumps/hybrid.go
```diff
@@ -128,8 +130,25 @@ func (p *HybridPump) Init(config interface{}) error {
 			p.trackAllPaths = trackAllPaths.(bool)
 		}
 
-		if storeAnalyticPerMinute, ok := meta["store_analytics_per_minute"]; ok {
-			p.storeAnalyticPerMinute = storeAnalyticPerMinute.(bool)
+		if storeAnalyticsPerMinute, ok := meta["store_analytics_per_minute"].(bool); ok {
```
sredxny, line 133: I think we should consider here the use case in which `meta["store_analytics_per_minute"] `is nil, this will throw a panic when casting. Even if we think that this will never be nil we should validate it. Toughts?

pumps/hybrid.go
```diff
@@ -128,8 +130,25 @@ func (p *HybridPump) Init(config interface{}) error {
...
+			p.storeAnalyticsPerMinute = storeAnalyticsPerMinute
+			p.aggregationTime = 1
+		} else {
+			aggregationTime, ok := meta["aggregation_time"].(int)
```
sredxny, line 137: should we validate as well if this is nil?

pumps/hybrid.go
```diff
@@ -128,8 +130,25 @@ func (p *HybridPump) Init(config interface{}) error {
...
+			if !ok {
+				p.log.Warn("aggregation_time is not configured. The default value will be set to 60")
```
sredxny, line 139: lets put something like: `p.log.Warn("aggregation_time is not configured. The default value will be used (60 minutes)")`

or something similar

pumps/hybrid.go
```diff
@@ -128,8 +130,25 @@ func (p *HybridPump) Init(config interface{}) error {
...
+				p.aggregationTime = 60
+			} else {
+				if aggregationTime > 60 || aggregationTime < 1 {
+					p.log.Warn("aggregation_time range is from 1 to 60. The default value will be set to 60")
+					p.aggregationTime = 60
+				}
+				p.aggregationTime = aggregationTime
+			}
+		}
```
sredxny, line 148: I think we can replace this with:

```
aggregationTime, ok := meta["aggregation_time"].(int)

if aggregationTime > 60 || aggregationTime < 1 {
					p.log.Warnf("aggregation_time should be between 1 and 60, Found: %v. The default value will be used  (60 minutes)",aggregationTime)
					p.aggregationTime = 60
				}

```

pumps/mongo_aggregate.go
```diff
@@ -261,12 +288,30 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{})
...
 		// calculate aggregates
-		analyticsPerOrg := analytics.AggregateData(data, m.dbConf.TrackAllPaths, m.dbConf.IgnoreTagPrefixList, m.dbConf.StoreAnalyticsPerMinute, true)
-
+		analyticsPerOrg := analytics.AggregateData(data, m.dbConf.TrackAllPaths, m.dbConf.IgnoreTagPrefixList, m.dbConf.MongoURL, m.dbConf.AggregationTime, true)
 		// put aggregated data into MongoDB
 		for orgID, filteredData := range analyticsPerOrg {
 			err := m.DoAggregatedWriting(ctx, orgID, filteredData)
 			if err != nil {
+				// checking if the error is related to the document size and AggregateSelfHealing is enabled
+				if strings.Contains(err.Error(), "Size must be between 0 and 16793600(16MB)") && m.dbConf.EnableAggregateSelfHealing {
```
sredxny, line 297: Would be the same error for the any supported DB that the customer use? Am wondering if this would be the same for cosmo,mongo, documentDB or should we isolate this error into its own function ?

pumps/mongo_aggregate.go
```diff
@@ -261,12 +288,30 @@ func (m *MongoAggregatePump) WriteData(ctx context.Context, data []interface{})
...
+					// if the AggregationTime setting is already set to 1, we can't do anything else
+					if m.dbConf.AggregationTime == 1 {
+						m.log.Warning("AggregationTime is equal to 1 minute, unable to reduce it further. Skipping self-healing.")
+						return err
+					}
```
sredxny, line 302: Even if the value is set to 1, shouldn't we create a new document?
We cannot divide the aggregationTime, but makes sense to create a new document anyway?

pumps/mongo_aggregate.go
```diff
@@ -416,3 +460,25 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
 func (m *MongoAggregatePump) WriteUptimeData(data []interface{}) {
 	m.log.Warning("Mongo Aggregate should not be writing uptime data!")
 }
+
+// getLastDocumentTimestamp will return the timestamp of the last document in the collection
+func getLastDocumentTimestamp(session *mgo.Session, collectionName string) (time.Time, error) {
+	var doc bson.M
+	err := session.DB("").C(collectionName).Find(nil).Sort("-$natural").One(&doc)
```
sredxny, line 467: will this work in all the databases?

sredxny (changes requested):

pumps/mongo_aggregate.go
tbuchaillot: good point @sredxny. @mativm02 did you have a look at CosmoDB/DocDB?

pumps/mongo_aggregate.go
```diff
@@ -416,3 +460,25 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
...
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	return doc["timestamp"].(time.Time), nil
```
tbuchaillot, line 472: I think it's unsafe to return something casted like this. Please do the validation before.

pumps/mongo_aggregate.go
tbuchaillot: I don't agree with this one @sredxny - If we are in the minimum aggregation time, trying to write records and still hitting the max MB, I think we should error

pumps/mongo_aggregate.go
tbuchaillot, line 297: @mativm02 Let's move this to a separated function

pumps/mongo_aggregate.go
```diff
@@ -170,12 +177,32 @@ func (m *MongoAggregatePump) Init(config interface{}) error {
 	if m.dbConf.ThresholdLenTagList == 0 {
 		m.dbConf.ThresholdLenTagList = THRESHOLD_LEN_TAG_LIST
 	}
+	// if StoreAnalyticsPerMinute is set to true, the aggregation time will be set to 1.
+	// if not, the aggregation time will be set to the value of the field AggregationTime.
+	// if there is no value for AggregationTime, it will be set to 60.
+	if m.dbConf.StoreAnalyticsPerMinute {
```
tbuchaillot, line 183: @mativm02 can we encapsulate this into a separated function and add tests? it's going to be easier to test.

pumps/mongo_aggregate.go
```diff
@@ -170,12 +177,32 @@ func (m *MongoAggregatePump) Init(config interface{}) error {
...
 
 	m.log.Debug("MongoDB DB CS: ", m.dbConf.GetBlurredURL())
 	m.log.Info(m.GetName() + " Initialized")
 
+	// look for the last record timestamp stored in the collection
+	lastTimestampAgggregateRecord, err := getLastDocumentTimestamp(m.dbSession, analytics.AgggregateMixedCollectionName)
+
+	// we will set it to the lastDocumentTimestamp map to track the timestamp of different documents of different Mongo Aggregators
+	if err != nil {
+		m.log.Warn("Unable to get last timestamp from aggregate collection: ", err)
```
tbuchaillot, line 201: @mativm02 this Warn is going to show every time we try to cold-start pump since the collection probably is not going to exist.
I'd change the message and the type of the log to Debug.

analytics/aggregate.go
```diff
@@ -20,6 +20,9 @@ const (
 	AggregateSQLTable             = "tyk_aggregated"
 )
 
+// lastDocumentTimestamp is a map to store the last document timestamps of different Mongo Aggregators
+var lastDocumentTimestamp = make(map[string]time.Time)
```
tbuchaillot, line 24: @mativm02 why is this a global variable? What would happen if we have multiple mongo-aggregate and sql-aggregate pumps running at the same time?

pumps/sql_aggregate.go
```diff
@@ -152,7 +150,15 @@ func (c *SQLAggregatePump) WriteData(ctx context.Context, data []interface{}) er
...
-		analyticsPerOrg := analytics.AggregateData(data[startIndex:endIndex], c.SQLConf.TrackAllPaths, c.SQLConf.IgnoreTagPrefixList, c.SQLConf.StoreAnalyticsPerMinute, false)
+		// if StoreAnalyticsPerMinute is set to true, we will create new documents with records every 1 minute
+		var aggregationTime int
+		if c.SQLConf.StoreAnalyticsPerMinute {
+			aggregationTime = 1
+		} else {
+			aggregationTime = 60
+		}
+
+		analyticsPerOrg := analytics.AggregateData(data[startIndex:endIndex], c.SQLConf.TrackAllPaths, c.SQLConf.IgnoreTagPrefixList, "", aggregationTime, false)
```
tbuchaillot, line 161: @mativm02 why the dbIdentifier is empty here?

tbuchaillot (changes requested):

pumps/hybrid.go
mativm02: Do you mean first checking the value of "ok" and then, setting a type for storeAnalyticsPerMinute? something like this:
`	if storeAnalyticsPerMinute, ok := meta["store_analytics_per_minute"]; ok {
			p.storeAnalyticsPerMinute = storeAnalyticsPerMinute.(bool)
			p.aggregationTime = 1
		}`

mativm02: I think we are validating it through the "ok" variable, am I wrong? 
When store_analytics_per_minute is nil (it does not exist in the map), "ok" will be equal to false, and we're handling that situation. Let me know if there is something I am not realizing!

pumps/hybrid.go
mativm02: Nice! It looks cleaner. But is it right not to check the value of "ok"?

pumps/hybrid.go
mativm02: Same as above. Checking the "ok" value is not a way to validate if aggregationTime is nil?

pumps/mongo_aggregate.go
mativm02: Oops, I checked with CosmosDB and it throws a different error! Will find a way to handle this situation.

pumps/mongo_aggregate.go
mativm02: Hmm. If the AggregationTime is already set to 1, it means that we're creating new documents every 1 minute. So if you've created a document at 10:46 and it hits 16mb before 10:47, we can't create a new document, because the time is still 10:46, right? If we want to handle these situations, we should include seconds (i.e: creating documents every 30 seconds).
If the document hits 16mb, but now the time is 10:47, there won't be any problem because a new document will be created automatically.
Let me know if there is something I am not seeing!

pumps/mongo_aggregate.go
mativm02: We can encapsulate it into a function (sending a pointer to MongoAggregatePump as a parameter) or into a *MongoAggregatePump method. Which one would be the best way?

analytics/aggregate.go
mativm02: @tbuchaillot because the function `AggregateData` is being used by Mongo and SQL. I could add a `lastDocumentTimestamp` field to `MongoAggregateConf` for example, and send it to to `AggregateData` through a parameter in order to get the last timestamp of Mongo instance.
The problem I noticed is that SQL wouldn't have this field in their config struct. That's why I've created this variable as global. We may think a better way to store this data.

The idea of the map is to organise every Mongo instance by their MongoURL (that's the key of the map), and the value is the last aggregation timestamp. So if we have multiple Mongo instances running, all of them we'll be stored in the map by their MongoURL. 
This feature is not being applied to SQL yet.

pumps/sql_aggregate.go
mativm02: @tbuchaillot because dbIdentifier is being used as the key of the map `lastDocumentTimestamp` we were talking before. This map is only being used for mongo. If the string is empty, it means the database != mongo (it's SQL in this case).
We can think about a cleaner way to handle this. Maybe have different functions to set timestamp to new data added to documents for every database service? Instead of handling all of them in one function. After setting the timestamp, we can continue using `AggregateData`.

pumps/hybrid.go
sredxny: done!

pumps/hybrid.go
sredxny: if ok is false, then `aggregationTime` will be zero, which is less than 1, hence it will enter into the conditional right?

sredxny:
Support for all the databases is still missing. So, this is a work in progress

analytics/aggregate.go
tbuchaillot: @mativm02 my point here is that you can have multiple pumps running and aggregating data at the same time.  Using a global var with concurrent writes can have some problems so I'd suggest switching to a sync.Map or use mutexes here.

tbuchaillot: Also, for SQL cases maybe it's worth using the connection string? Imagine that you have 2 SQL aggregate pumps writing in different dbs.

pumps/mongo_aggregate.go
tbuchaillot: @mativm02 can you check @sredxny question?

pumps/mongo_aggregate.go
```diff
@@ -416,3 +443,69 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
...
+
+// divideAggregationTime divides by two the analytics stored per minute setting
+func (m *MongoAggregatePump) divideAggregationTime() {
+	if m.dbConf.AggregationTime == 1 {
+		m.log.Warn("Analytics Stored Per Minute is set to 1, unable to divide")
+		return
+	}
+	oldAggTime := m.dbConf.AggregationTime
+	m.dbConf.AggregationTime /= 2
+	m.log.Info("Analytics Stored Per Minute dicreased from ", oldAggTime, " to ", m.dbConf.AggregationTime)
```
tbuchaillot, line 468: I'd use a Warn log here.

pumps/mongo_aggregate.go
```diff
@@ -416,3 +443,69 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
...
+	if ts, ok := doc["timestamp"].(time.Time); ok {
+		return ts, nil
+	}
+	return time.Time{}, errors.New("timestamp of type: time.Time not found in bson map")
+}
+
+// divideAggregationTime divides by two the analytics stored per minute setting
+func (m *MongoAggregatePump) divideAggregationTime() {
+	if m.dbConf.AggregationTime == 1 {
+		m.log.Warn("Analytics Stored Per Minute is set to 1, unable to divide")
```
tbuchaillot, line 463: I'd use a debug log here.

pumps/mongo_aggregate.go
```diff
@@ -416,3 +443,69 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
...
+	m.log.Info("Analytics Stored Per Minute dicreased from ", oldAggTime, " to ", m.dbConf.AggregationTime)
+}
+
+// ShouldSelfHeal returns true if the pump should self heal
+func (m *MongoAggregatePump) ShouldSelfHeal(err error) bool {
+
+	const StandarMongoSizeError = "Size must be between 0 and 16793600(16MB)"
+	const CosmosSizeError = "Request size is too large"
+
+	fmt.Println("-----------should self heal---------------------")
```
tbuchaillot, line 477: Please remove the printlns

pumps/mongo_aggregate.go
```diff
@@ -416,3 +443,69 @@ func (m *MongoAggregatePump) collectionExists(name string) (bool, error) {
...
+	oldAggTime := m.dbConf.AggregationTime
+	m.dbConf.AggregationTime /= 2
+	m.log.Info("Analytics Stored Per Minute dicreased from ", oldAggTime, " to ", m.dbConf.AggregationTime)
+}
+
+// ShouldSelfHeal returns true if the pump should self heal
+func (m *MongoAggregatePump) ShouldSelfHeal(err error) bool {
+
+	const StandarMongoSizeError = "Size must be between 0 and 16793600(16MB)"
+	const CosmosSizeError = "Request size is too large"
```
tbuchaillot, line 475: @mativm02 what about docdb?

pumps/mongo_aggregate_test.go
```diff
@@ -101,7 +101,7 @@ func TestDoAggregatedWritingWithIgnoredAggregations(t *testing.T) {
 			//we build the query using the timestamp as we do in aggregated analytics
 			query := bson.M{
 				"orgid":     "123",
-				"timestamp": time.Date(timeNow.Year(), timeNow.Month(), timeNow.Day(), timeNow.Hour(), 0, 0, 0, timeNow.Location()),
+				"timestamp": time.Date(timeNow.Year(), timeNow.Month(), timeNow.Day(), timeNow.Hour(), timeNow.Minute(), 0, 0, timeNow.Location()),
```
tbuchaillot, line 104: @mativm02 why are we changing this test?

pumps/mongo_aggregate_test.go
```diff
@@ -125,3 +125,113 @@ func TestDoAggregatedWritingWithIgnoredAggregations(t *testing.T) {
...
+			}
+
+			collectionName := analytics.AgggregateMixedCollectionName
+
+			thisSession := pmp1.dbSession.Copy()
+			defer thisSession.Close()
+
+			analyticsCollection := thisSession.DB("").C(collectionName)
+
+			// we build the query using the timestamp as we do in aggregated analytics
```
tbuchaillot, line 215: @mativm02 you are not using the timestamp in the query 😛

pumps/sql_aggregate.go
tbuchaillot: @mativm02 What would happen if you have 2 SQL Aggregate pumps running and writing data to different dbs?

tbuchaillot (changes requested):

pumps/mongo_aggregate_test.go
mativm02: Let me explain it with an example:
What was happening before is that if the aggregated time was set to 60 minutes, and the current time was 10:13 am, the document timestamp would be set to 10:00. The timestamp of the next document would be 11:00 am. 
Now, if the aggregated time is 60 minutes and the current time is 10:13 am, the document timestamp will be set to 10:13 am. The timestamp of the next document will be 11:13 am.

tbuchaillot: @mativm02 by default, we should maintain the existing behavior, so we don't break backward compatibility with the current pump users when they upgrade the version

mativm02: @tbuchaillot okay! now if the aggregation time is set to 60, it will create a new document every o'clock hour

pumps/mongo_aggregate_test.go
mativm02: I forgot to modify the comment

pumps/mongo_aggregate.go
mativm02: added in the last commit!

pumps/mongo_aggregate.go
mativm02: It's currently working on Cosmos, DocDB and Standard Mongo!

pumps/sql_aggregate.go
mativm02: I will test that use case, but nothing should happen, SQL Aggregate should work as always 🤔

tbuchaillot (changes requested):
@mativm02 First, can you document this change?

Also, what would happen if a pump is restarted? How's the behavior with two tyk-pumps writing over the same collection?

analytics/aggregate.go
```diff
@@ -559,11 +566,35 @@ func AggregateData(data []interface{}, trackAllPaths bool, ignoreTagPrefixList [
 
 			// Set the hourly timestamp & expiry
 			asTime := thisV.TimeStamp
-			if storeAnalyticPerMinute {
-				thisAggregate.TimeStamp = time.Date(asTime.Year(), asTime.Month(), asTime.Day(), asTime.Hour(), asTime.Minute(), 0, 0, asTime.Location())
-			} else {
+			// get the last document timestamp
+			lastDocumentTS, ok := getLastDocumentTimestamp(dbIdentifier)
```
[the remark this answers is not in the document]
mativm02: done!

sredxny (approved):
lgtm

tbuchaillot (approved):

PR #489, opened 2022-09-13
[TT-6012] Implement New Graph Granular Mongo Pump

kofoworola:
## Description
This Pr brings the introduction of a new mongo pump for graphql data specifically
See:
[what we are logging and why we need it](https://tyktech.atlassian.net/wiki/spaces/EN/pages/1724481598/What+do+we+need+to+log+for+GQL)
[storage format](https://tyktech.atlassian.net/wiki/spaces/EN/pages/1771470898/TT-6119+Graph+Analytics-How+are+we+storing+data)

The configuration for the new pump is similar to the existing graph pump:
```
"mongo-graph": {
      "type": "mongo-graph",
      "meta": {
        "collection_name": "graph_analytics",
        "mongo_url": "mongodb://localhost:27017/tyk_analytics"
      }
    }
```

## Related Issue

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


tbuchaillot (changes requested):
Please address linter comments and test coverage of `graph_record.go`

analytics/analytics.go
```diff
@@ -22,6 +22,13 @@ const (
 	PredefinedTagGraphAnalytics = "tyk-graph-analytics"
 )
 
+const (
```
Aenimus, line 25: It might make sense for pump to have its own version, but I just wanted to let you know that this exact thing exists in graphql-go-tools.

analytics/analytics.go
```diff
@@ -22,6 +22,13 @@ const (
...
+	operationTypeQuery        = "query"
+	operationTypeMutation     = "mutation"
+	operationTypeSubscription = "subscription"
+	operationTypeUnknown      = "unknown"
+)
```
Aenimus, line 30: It might make sense for pump to have its own version, but I just wanted to let you know that this exact thing exists in graphql-go-tools.

pumps/mongo.go
```diff
@@ -553,7 +552,10 @@ func (m *MongoPump) AccumulateSet(data []interface{}) [][]interface{} {
 		}
 
 		// Skip this record if it is a graph analytics record, they will be handled in a different pump
-		if thisItem.IsGraphRecord() {
+		if thisItem.IsGraphRecord() && !forGraph {
```
Aenimus, line 555: You can simplify this to
`if thisItem.IsGraphRecord() != forGraph {
    continue
}`

Aenimus: Or XOR ( a ^ !b) if you want to be fancy and explicit.

analytics/analytics.go
kofoworola: Oh, didn't know about that, i'll look through the go-tools and use it's version, it's better if there is some standard followed.

pumps/mongo.go
kofoworola: Yeah, imo while it is simplified, I wanted to be as verbose as possible to aid understanding.

Aenimus: Honestly, I read this, and all I could think was "why not just compare the two not to be equal?". That made it more confusing, at least for me.

If you really think it's better, I'd at least have the variables in the same order.

Aenimus: @buraksezer could you please weigh in?

buraksezer: I think `forGraph` is bad here. It makes it harder to understand. We can reorganize `AccumulateSet` function. We can add a new function called `AccumulateSetForGraph` and it can share code with `AccumulateSet`. They call a common function called `accumulateSetCommon` or something?

kofoworola: @buraksezer so we can't split it cause it's the exact same functionality, only thing is the `forGraph` determins if only graph records will be accumulated or only non graph record. I think @Aenimus solution is better, with a better named variable

Aenimus: Maybe we could also rename the variable also?

kofoworola: Yup done that.

pumps/graph_mongo.go
```diff
@@ -0,0 +1,145 @@
...
+		err = mapstructure.Decode(config, &g.dbConf.BaseMongoConf)
+		g.log.WithFields(logrus.Fields{
+			"url":             g.dbConf.GetBlurredURL(),
+			"collection_name": g.dbConf.CollectionName,
+		}).Info("Init")
+		if err != nil {
+			panic(g.dbConf.BaseMongoConf)
+		}
+	}
+	if err != nil {
```
buraksezer, line 47: Checking an error and breaking the flow if it is a non-nil look more idiomatic. 

```go
if err != nil {
   return fmt.Errorf("error while decode configuration: %w", err)
}

err = mapstructure.Decode(config, &g.dbConf.BaseMongoConf)
g.log.WithFields(logrus.Fields{
	"url":             g.dbConf.GetBlurredURL(),
	"collection_name": g.dbConf.CollectionName,
}).Info("Init")
if err != nil {
	panic(g.dbConf.BaseMongoConf)
}
```

Caller of `Init` should log this error, I think.

pumps/graph_mongo.go
```diff
@@ -0,0 +1,145 @@
...
+		g.dbConf.MaxDocumentSizeBytes = 10 * MiB
+	}
+
+	g.connect()
+
+	g.capCollection()
+
+	indexCreateErr := g.ensureIndexes()
+	if indexCreateErr != nil {
+		g.log.Error(indexCreateErr)
```
buraksezer, line 67: Is it safe to move on from that point? If index creation fails, it may hurt the performance in production env.

pumps/graph_mongo.go
```diff
@@ -0,0 +1,145 @@
...
+
+	g.log.Info(g.GetName() + " Initialized")
+
+	return nil
+}
+
+func (g *GraphMongoPump) WriteData(ctx context.Context, data []interface{}) error {
+	collectionName := g.dbConf.CollectionName
+	if collectionName == "" {
+		g.log.Fatal("No collection name!")
```
buraksezer, line 81: `WriteData` function has an error type as a return parameter. Does the caller of `WriteData` method log this error? Returning an error if the collection name is empty looks more idiomatic to me.

pumps/graph_mongo.go
```diff
@@ -0,0 +1,145 @@
...
+			// make a graph record array with variable length in case there are errors with some conversion
+			finalSet := make([]interface{}, 0)
+			for _, d := range dataSet {
+				r, ok := d.(analytics.AnalyticsRecord)
+				if !ok {
+					continue
+				}
+				gr, err := r.ToGraphRecord()
+				if err != nil {
+					g.log.WithError(err).Warn("error converting 1 record to graph record")
```
buraksezer, line 107: Missing `continue` statement?

pumps/graph_mongo.go
```diff
@@ -0,0 +1,145 @@
+package pumps
```
buraksezer, line 1: How do we test the business logic in `graph_mongo.go`?

analytics/graph_record.go
```diff
@@ -0,0 +1,318 @@
...
+func getOperationSelectionFieldDefinition(operationType ast.OperationType, opSelectionName string, schema *ast.Document) (int, error) {
+	var (
+		node  ast.Node
+		found bool
+	)
+	switch operationType {
+	case ast.OperationTypeQuery:
+		node, found = schema.Index.FirstNodeByNameBytes(schema.Index.QueryTypeName)
+		if !found {
+			return -1, fmt.Errorf("missing query type declaration")
```
buraksezer, line 286: I think it is a good idea to return `ast.InvalidRef` instead of `-1`.

analytics/graph_record.go
```diff
@@ -0,0 +1,318 @@
...
+	for _, selRef := range req.SelectionSets[field.SelectionSet].SelectionRefs {
+		sel := req.Selections[selRef]
+		if sel.Kind != ast.SelectionKindField {
+			continue
+		}
+		fieldListForType = append(fieldListForType, req.FieldNameString(sel.Ref))
+
+		// get the field definition and run this function on it
+		fieldDefRef := getObjectFieldRefWithName(req.FieldNameString(sel.Ref), typeDef, schema)
+		if fieldDefRef == -1 {
```
buraksezer, line 173: I think it is a good idea to return `ast.InvalidRef` instead of `-1`.

pumps/graph_mongo.go
kofoworola: AFAIK it was the intended behaviour for the other mongo based pumps

pumps/graph_mongo.go
kofoworola: Yeah i initially copied the current mongo implementation but changed the structure and logic to return an error rather than panic, the caller logs this error and skips the pump initialization

analytics/graph_record.go
```diff
@@ -0,0 +1,317 @@
...
+		typeRef := schema.ResolveUnderlyingType(schema.FieldDefinitions[selFieldDefRef].Type)
+		if schema.TypeIsScalar(typeRef, schema) || schema.TypeIsEnum(typeRef, schema) {
+			continue
+		}
+		fieldTypeMap[sel.Ref] = getObjectTypeRefWithName(schema.TypeNameString(typeRef), schema)
+	}
+	return fieldTypeMap, nil
+}
+
+func recursivelyExtractTypesAndFields(fieldRef, typeDef int, resp map[string][]string, req, schema *ast.Document) {
```
tbuchaillot, line 157: @kofoworola why recursively in the name of the func? Can we add tests to this?

analytics/graph_record.go
```diff
@@ -0,0 +1,317 @@
...
+		recursivelyExtractTypesAndFields(fieldRef, typeDefRef, typesToFieldsMap, request, schema)
+	}
+	record.Types = typesToFieldsMap
+
+	// get errors
+	responseDecoded, err := base64.StdEncoding.DecodeString(a.RawResponse)
+	if err != nil {
+		return record, nil
+	}
+	resp, err := http.ReadResponse(bufio.NewReader(bytes.NewReader(responseDecoded)), nil)
```
tbuchaillot, line 97: @kofoworola why are we doing this? only to see if there's an error in the response body?

analytics/graph_record.go
```diff
@@ -0,0 +1,317 @@
...
+
+	AnalyticsRecord `bson:",inline"`
+
+	OperationType string
+	Variables     string
+	Errors        []graphError
+	HasErrors     bool
+}
+
+func (a *AnalyticsRecord) ToGraphRecord() (GraphRecord, error) {
```
tbuchaillot, line 32: @kofoworola did you benchmark how much resource/time all this processing adds?

tbuchaillot (changes requested):

pumps/graph_mongo.go
kofoworola: Updated the tests

pumps/graph_mongo.go
```diff
@@ -0,0 +1,146 @@
...
+
+	g.log.Debug("MongoDB DB CS: ", g.dbConf.GetBlurredURL())
+	g.log.Debug("MongoDB Col: ", g.dbConf.CollectionName)
+
+	g.log.Info(g.GetName() + " Initialized")
+
+	return nil
+}
+
+func (g *GraphMongoPump) WriteData(ctx context.Context, data []interface{}) error {
```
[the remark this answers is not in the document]
kofoworola: YEs added the tests

analytics/graph_record.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// getOperationSelectionFieldDefinition gets the schema's field definition ref for the selection set of the operation type in question
+func getOperationSelectionFieldDefinition(operationType ast.OperationType, opSelectionName string, schema *ast.Document) (int, error) {
+	var (
+		node  ast.Node
+		found bool
+	)
+	switch operationType {
+	case ast.OperationTypeQuery:
```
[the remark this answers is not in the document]
kofoworola: Yes

analytics/graph_record.go
kofoworola: Yes we need to read the response body properly to check for errors in the response so i used http's native read response to accommodate for different chunked response lengths etc

analytics/graph_record.go
```diff
@@ -0,0 +1,317 @@
...
+	if !ok {
+		return ast.InvalidRef
+	}
+	if n.Kind != ast.NodeKindObjectTypeDefinition {
+		return ast.InvalidRef
+	}
+	return n.Ref
+}
+
+func generateNormalizedDocuments(requestRaw, schemaRaw []byte) (r, s *ast.Document, operationName string, err error) {
```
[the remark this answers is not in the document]
kofoworola: These functions were split from the general ToGraphRecord function, the unit tests for the graph_record cover these functions and breakages in these functions

tbuchaillot (changes requested):
Looks good overall! added some small comments @kofoworola.

Also, please add documentation in readme specifying this will be available from pump 1.7+ :)

pumps/graph_mongo_test.go
```diff
@@ -0,0 +1,323 @@
...
+		MongoPump: MongoPump{
+			dbConf: &conf,
+		},
+	}
+	pump.log = log.WithField("prefix", mongoPrefix)
+	pump.MongoPump.CommonPumpConfig = pump.CommonPumpConfig
+	pump.dbConf.CollectionCapEnable = true
+	pump.dbConf.CollectionCapMaxSizeBytes = 0
+
+	type customRecord struct {
```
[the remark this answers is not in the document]
kofoworola: Oh yeah this was from a previous iteration of test code, i'd modify it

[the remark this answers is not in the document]
kofoworola: Oh, cause it's not a analytics record, it just holds information (per test case) that is modified and added to a dummmy analytics record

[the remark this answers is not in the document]
tbuchaillot: oh I see. Can we add different status_code to the test case/ analytic records?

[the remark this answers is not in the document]
kofoworola: Yup added, thanks for pointing that, also updated the documentation.

tbuchaillot (approved):
LGTM @kofoworola 🚀

Issue #491, opened 2022-09-14
RFC: Prometheus: `tyk_http_status_per_path` cardinality issues leading to scrape timeouts.


hsyed-dojo:
Hello Folks,

I'm looking at contributing a tactical fix for our API gateway team. Completely new to Tyk so apologies if am overlooking existing workarounds.

The `path` field in `tyk_http_status_per_path` is generating too much cardinality which in turn is timing out our scrape jobs. We'd like to exclude this particular metric family from exposition entirely.

I propose adding a `DisabledMetrics []string` (name?) to `PrometheusConf` and any such metrics families get their `PrometheusMetric.enabled` flipped to false. 

If this sounds good I'm happy to submit a PR.

tbuchaillot:
Hey @hsyed-dojo! 

It sounds good indeed! Please rise a PR and add me as reviewer 👍

PR #492, opened 2022-09-15
TT-6799 Prometheus pump: Support to disable certain metric families from exposition. 

hsyed-dojo:
(cc: @tbuchaillot)

See related issue

## Description
See related issue

## Related Issue
#491 
https://tyktech.atlassian.net/browse/TT-6799
## Motivation and Context
See related issue

## How This Has Been Tested
Added some tests for the simple change, relying on a review for the rest.

## Types of changes

## Checklist

- [x] All new and existing tests passed. (I just tested the Prometheus tests are passsing because no local testing tooling). 
  - [x] `go fmt -s`
  - [x] `go vet`


pumps/prometheus.go
```diff
@@ -69,7 +71,7 @@ type PrometheusMetric struct {
 	aggregatedObservations bool
 }
 
-//histogramCounter is a helper struct to mantain the totalRequestTime and hits in memory
+// histogramCounter is a helper struct to mantain the totalRequestTime and hits in memory
```
hsyed-dojo, line 74: sorry for the formatting changes. Intellij formats file on save

pumps/prometheus.go
```diff
@@ -160,30 +162,55 @@ func (p *PrometheusPump) Init(conf interface{}) error {
 	}
 
 	//first we init the base metrics
-	for _, metric := range p.allMetrics {
```
hsyed-dojo, line 163: split this out into its own function to be able to easily test

pumps/prometheus_test.go
```diff
@@ -2,14 +2,16 @@ package pumps
...
 	"testing"
 
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestInitVec(t *testing.T) {
+func TestPrometheusInitVec(t *testing.T) {
```
hsyed-dojo, line 14: Updated test names in this file to have `TestPrometheus` prefix to easily be able to select tests by regex.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Initial connector which outputs all analytics into a single elasticsearch index
 - Enabled the ES connecter to save all output to a rolling index rather than a static index
+- Add support to the Prometheus pump to congigure metrics families from exposition
```
joshblakeley, line 7: Export would be more natural english than exposition. Thanks!

pumps/prometheus.go
```diff
@@ -160,30 +162,55 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
 }
 
-//InitCustomMetrics initialise custom prometheus metrics based on p.conf.CustomMetrics and add them into p.allMetrics
+func (p *PrometheusPump) disableExcludedMetrics() []string {
+	//configure base or custom metrics as disabled if needed. This disables exposition entirely during scrapes.
+	var disabledMetrics []string
+	{
+		toDisableSet := map[string]struct{}{}
+		for _, metric := range p.conf.DisabledMetrics {
+			toDisableSet[metric] = struct{}{}
```
hsyed-dojo, line 190: wish the sort package had a `Contains` method :/

CHANGELOG.md
hsyed-dojo: I picked up the term from OpenTelemetry/[OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md) documentation. Exposition is the term used to describe a particular scrape cycle in the prometheus model.

Happy to change though.

joshblakeley: Very interesting thanks for the context. It looks like that term is used widely enough so happy to keep it.

pumps/prometheus_test.go
tbuchaillot: makes sense @hsyed-dojo !

pumps/prometheus.go
```diff
@@ -160,30 +162,55 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
+				metric.enabled = false
+				disabledMetrics = append(disabledMetrics, metric.Name)
+			}
+		}
+	}
+	return disabledMetrics
+}
+
+func (p *PrometheusPump) initBaseMetrics() {
+	for _, metric := range p.allMetrics {
```
tbuchaillot, line 204: @hsyed-dojo why don't we check and disable the metrics directly inside this method and before initialization? It doesn't make a lot of sense init the metric if we are disabling it right after IMO.

tbuchaillot (changes requested):

pumps/prometheus.go
hsyed-dojo: The approach I took also applies to any custom metrics that might be configured.

@tbuchaillot The way it is currently implemented we have support to disable any custom metrics that might have been configured. This could be useful for temporarily disabling a metric. 

WDYT ? Happy to apply your suggestion if that is preferable.

sredxny: yes, for sure we have to be clear and avoid any point of confusion. Why not only remove the metric from the config file? am I missing something?

pumps/prometheus.go
```diff
@@ -178,12 +174,39 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
+	p.log.WithField("disabled_metrics", disabledMetrics).Info(p.GetName() + " Initialized")
 
 	return nil
 }
 
-//InitCustomMetrics initialise custom prometheus metrics based on p.conf.CustomMetrics and add them into p.allMetrics
+func (p *PrometheusPump) initBaseMetrics() []string {
+	for _, metric := range p.allMetrics {
+		metric.aggregatedObservations = p.conf.AggregateObservations
+		errInit := metric.InitVec()
```
tbuchaillot, line 185: @hsyed-dojo I think it's going to be better to create the `toDisableSet` first, and check if we should `errInit := metric.InitVec()` based on the metric name here. In this way, we avoid initializing the metric if it's already disabled and avoid the need of unregistering it from Prometheus client. 
Maybe we can create a `PrometheusMetric` method for this and make this cleaner / easier to test.   
What do you think?

pumps/prometheus.go
```diff
@@ -178,12 +174,39 @@ func (p *PrometheusPump) Init(conf interface{}) error {
...
 		log.Fatal(http.ListenAndServe(p.conf.Addr, nil))
 	}()
-	p.log.Info(p.GetName() + " Initialized")
+	p.log.WithField("disabled_metrics", disabledMetrics).Info(p.GetName() + " Initialized")
 
 	return nil
 }
 
-//InitCustomMetrics initialise custom prometheus metrics based on p.conf.CustomMetrics and add them into p.allMetrics
+func (p *PrometheusPump) initBaseMetrics() []string {
```
tbuchaillot, line 182: @hsyed-dojo I don't think we should return the disabled metrics here, it doesn't make sense that the initBaseMetrics func return that IMO.

tbuchaillot (changes requested):

pumps/prometheus.go
hsyed-dojo: If one ops engineer goes and disables a metric and another ops engineer expects that metric to still be active in an environment I suspect it would help them to see a log message indicating that the metric was disabled. I guess they can inspect the config 🤷‍♂️ .

Will remove it.

pumps/prometheus.go
hsyed-dojo: @tbuchaillot that is the approach I would have taken if the data flow did not already have the concept of disabling a metric during exposition. Having two codepaths to turn off metric exposition seems less clear to reason about.

Happy to update. Let me know.

tbuchaillot: @hsyed-dojo the current concept of disabling metrics is only for initialization failure. In fact, we are enabling when init correctly instead of disabling them IIRC.
If you want to give it a try to this approach, it would be amazing :)

hsyed-dojo: @tbuchaillot done

tbuchaillot (approved):
LGTM @hsyed-dojo ! please resolve the conflict so we can merge it 🚀

hsyed-dojo:
@tbuchaillot done, many thanks for the review!

tbuchaillot:
Thanks for the contribution @hsyed-dojo 🎉  This will be released in Pump 1.7 ( early December ).

tbuchaillot:
Hey @hsyed-dojo! This was just released in [Pump 1.7.0](https://github.com/TykTechnologies/tyk-pump/releases/tag/v1.7.0).
Thanks a lot for your contribution!

PR #502, opened 2022-10-24
TT-3067 - Add ability to use ssl_insecure_skip_verify with Elasticsearch Pump

mativm02:
## Description

Adding users the ability to use an insecure SSL (an autogenerated one), by skipping the SSL verification. We are also allowing the user to select cert and key files.
They can also set the option UseSSL to false in order to skip this process. 
## Related Issue

https://tyktech.atlassian.net/browse/TT-3067?atlOrigin=eyJpIjoiZDNjOTRkNTIxZmI0NDU3MDhlMWYzM2ZkYjQ1ZGZlZWYiLCJwIjoiaiJ9
## How This Has Been Tested

This has been tested by deploying a local Elastic Search with an autogenerated SSL certificate. I have also been using `--demo` flag in the pump to simulate requests.
[This is the tutorial I've used to deploy it with Docker.](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/configuring-tls-docker.html)

This is the exact command I ran to deploy pump:
`go build && TYK_PMP_PUMPS_ELASTICSEARCH_TYPE=elasticsearch TYK_PMP_PUMPS_ELASTICSEARCH_META_INDEXNAME=tyk_analytics TYK_PMP_PUMPS_ELASTICSEARCH_META_ELASTICSEARCHURL=https://localhost:9200 TYK_PMP_PUMPS_ELASTICSEARCH_META_ENABLESNIFFING=false TYK_PMP_PUMPS_ELASTICSEARCH_META_DOCUMENTTYPE=tyk_analytics TYK_PMP_PUMPS_ELASTICSEARCH_META_ROLLINGINDEX=false TYK_PMP_PUMPS_ELASTICSEARCH_META_EXTENDEDSTATISTICS=false TYK_PMP_PUMPS_ELASTICSEARCH_META_VERSION=7 TYK_PMP_PUMPS_ELASTICSEARCH_META_BULKCONFIG_WORKERS=2 TYK_PMP_PUMPS_ELASTICSEARCH_META_BULKCONFIG_FLUSHINTERVAL=60 TYK_PMP_PUMPS_ELASTICSEARCH_META_USESSL=true TYK_PMP_PUMPS_ELASTICSEARCH_META_SSLCERTFILE=/tmp/es01.crt TYK_PMP_PUMPS_ELASTICSEARCH_META_SSLKEYFILE=/tmp/es01.key TYK_PMP_PUMPS_ELASTICSEARCH_META_USERNAME=elastic TYK_PMP_PUMPS_ELASTICSEARCH_META_PASSWORD=password TYK_PMP_PUMPS_ELASTICSEARCH_META_SSLINSECURESKIPVERIFY=true
./tyk-pump --conf ../tyk-develop-env/confs/pump_elastic.conf --demo=634ed5f395014f77d8ec3e49 --demo-api=d79ecd1363784e6c7380cd58f8f44e8d
`

I also deployed an AWS ElasticSearch instance to test this feature using a valid certificate.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/elasticsearch.go
```diff
@@ -355,12 +370,50 @@ func (e *ElasticsearchPump) Init(config interface{}) error {
...
+	}
+
 	e.connect()
 
 	e.log.Info(e.GetName() + " Initialized")
 	return nil
 }
 
+// setTLSConfig sets the TLS config for the pump
+func (e *ElasticsearchPump) setTLSConfig() error {
```
tbuchaillot, line 387: @mativm02 Wouldn't be easier to create a helper function that receives SSLCertFile, SSLKeyFile, and SSLInsecureSkipVerify and returns a `&tls.Config`? IMO it's easier to test.

pumps/elasticsearch.go
```diff
@@ -355,12 +370,50 @@ func (e *ElasticsearchPump) Init(config interface{}) error {
 		e.log.Info("Index will have date appended to it in the format ", e.esConf.IndexName, "-YYYY.MM.DD")
 	}
 
+	if e.esConf.UseSSL {
```
tbuchaillot, line 373: @mativm02 why this is necessary on `Init` func instead of on `getOperator`?

tbuchaillot (changes requested):
@mativm02 added some comments. Please add some tests and the documentation of those new configs in the readme.

pumps/elasticsearch.go
mativm02: image
The requirement is asking to apply this logic in the Init function because if it fails, we would just stop the execution when it starts.

pumps/elasticsearch.go
mativm02: done!

mativm02:
talked with Andy and docs will be added in a separate PR :)

pumps/elasticsearch.go
```diff
@@ -71,6 +72,16 @@ type ElasticsearchConf struct {
...
+	// Enables SSL connection.
+	UseSSL bool `json:"use_ssl" mapstructure:"use_ssl"`
+	// Controls whether the pump client verifies the Elastic Search server's certificate chain and hostname.
+	SSLInsecureSkipVerify bool `json:"ssl_insecure_skip_verify" mapstructure:"ssl_insecure_skip_verify"`
+	// Can be used to set custom certificate file for authentication with Elastic Search.
+	SSLCertFile string `json:"ssl_cert_file" mapstructure:"ssl_cert_file"`
+	// Can be used to set custom key file for authentication with Elastic Search.
+	SSLKeyFile string `json:"ssl_key_file" mapstructure:"ssl_key_file"`
+
+	tlsConfig *tls.Config
```
sredxny, line 84: I would remove this field, as the value of this field depends on the others SSL fields then I don't think we need it. And instead build a function that returns the TLSConfig based on current configuration. Also consider that everytime that we add new fields in this kind of struct we need to set a json tag, Eg: `json:"-"` or `json:"tls_config"`

pumps/elasticsearch.go
```diff
@@ -143,6 +154,10 @@ func (e *ElasticsearchPump) getOperator() (ElasticsearchOperator, error) {
 		httpClient = &http.Client{Transport: &ApiKeyTransport{APIKey: conf.AuthAPIKey, APIKeyID: conf.AuthAPIKeyID}}
 	}
 
+	if conf.UseSSL {
+		httpClient = &http.Client{Transport: &http.Transport{TLSClientConfig: e.esConf.tlsConfig}}
```
sredxny, line 158: Instead of: `httpClient = &http.Client{Transport: &http.Transport{TLSClientConfig: e.esConf.tlsConfig}}` lets do something like:

`httpClient = &http.Client{Transport: &http.Transport{TLSClientConfig: e.esConf.getTLSConfig()}}`

pumps/elasticsearch.go
```diff
@@ -355,12 +370,53 @@ func (e *ElasticsearchPump) Init(config interface{}) error {
 		e.log.Info("Index will have date appended to it in the format ", e.esConf.IndexName, "-YYYY.MM.DD")
 	}
 
+	if e.esConf.UseSSL {
+		e.log.Debug("Getting SSL Certificates")
+		tlsConfig, err := setTLSConfig(e.esConf.SSLCertFile, e.esConf.SSLKeyFile, e.esConf.SSLInsecureSkipVerify)
+		if err != nil {
+			e.log.Fatal("Failed to set TLS config: ", err)
+		}
+		e.esConf.tlsConfig = tlsConfig
```
sredxny, line 379: if we remove the field, then this is not longer required

pumps/elasticsearch.go
```diff
@@ -355,12 +370,53 @@ func (e *ElasticsearchPump) Init(config interface{}) error {
...
+	}
+
 	e.connect()
 
 	e.log.Info(e.GetName() + " Initialized")
 	return nil
 }
 
+// setTLSConfig sets the TLS config for the pump
+func setTLSConfig(SSLCertFile, SSLKeyFile string, SSLInsecureSkipVerify bool) (*tls.Config, error) {
```
sredxny, line 389: This function is called `SetTLSConfig`, but it doesn't set anything. Instead let's rename it `GetTLSConfig`

sredxny (changes requested):

pumps/elasticsearch.go
mativm02: Understood - I'm removing it!

I didn't set the json tag as it's something that we're setting during the program execution, not from an env var. But makes totally sense to delete it. Thanks!

pumps/elasticsearch.go
mativm02: I'll do something similar because we need to check if GetTLSConfig() returns an error

pumps/elasticsearch.go
mativm02: I choose that name because it was setting the tlsConfig field of the esConf struct. But now that it's been deleted, I think `GetTLSConfig` is a better name for this function!

sredxny (approved):
lgtm

tbuchaillot (approved):

PR #504, opened 2022-10-25
[TT-6012] mongo graph records ignore max_document_size_bytes

kofoworola:
## Description

Added a fix where the `max_document_size_bytes` is ignored for graph reco

## Related Issue

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/mongo.go
```diff
@@ -552,21 +552,17 @@ func (m *MongoPump) AccumulateSet(data []interface{}, isForGraphRecords bool) []
...
-		if thisItem.IsGraphRecord() != isForGraphRecords {
+		isGraphRecord := thisItem.IsGraphRecord()
+		if isGraphRecord != isForGraphRecords {
 			continue
 		}
-		if isForGraphRecords {
-			if thisItem.RawRequest == "" || thisItem.RawResponse == "" || thisItem.ApiSchema == "" {
-				continue
-			}
-		}
```
buraksezer, line 562: Why did we remove this block? Does it break the `max-document-size-bytes` logic?

kofoworola: No it's an unneeded check, we should be discarding records that do not have those, that logic was initially there thinking the rawrequest had to absolutely be available, but now, we've realized it's possible to have records without them, in fact it's more than likely...The graph pump would throw a warning anyway when it encounters such records, it's better to be logged, than silently discarded

buraksezer (approved):

PR #507, opened 2022-10-27
TT-506/2 - Adding info about self-healing

mativm02:
## Description

Adding information about the changes made in https://github.com/TykTechnologies/tyk-pump/pull/486
## Related Issue

https://tyktech.atlassian.net/browse/TT-506?atlOrigin=eyJpIjoiN2EwMWU5MGZlODVmNGUxN2I4ZTNmNzJmNWNlMWNhNTAiLCJwIjoiaiJ9
## Motivation and Context

We need to describe what self-healing is and how it works. The documentation is being added in an external PR to not delay the PR that contains the code.
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
 TYK_PMP_PUMPS_MONGOAGG_META_USEMIXEDCOLLECTION=true
 ```
 
+###### Self Healing
+The default max size of a document in MongoDB is 16Mb. In Mongo Aggregate, when we try to update a document, and the size reaches the limit, an error is received.
+When Self Healing is enabled and a document size hits 16Mb, Pump will create a new document with the current timestamp and reduce the aggregation time configuration by half. So, if the current_aggregation time is equal to 50, it will be reduced to 25. Also, the data that was being written, is sent to the new document created using its timestamp. 
+Now, following this example, new data will be stored into this new document, and a new document will be created every 25 minutes.
+
+Note that if `store_analytics_per_minute` is equal to true, the value of `aggregation_time` will be equal to 1, no matter the value set. This was designed in this way because of backward compatibility.
+
 ###### Mongo Graph Pump
 As of Pump 1.7+, a new mongo is available called the `mongo_graph` pump. This pump is specifically for parsing 
 GraphQL and UDG requests, tracking information like types requested, fields requested, specific graphql body errors etc.
--- a/README.md
+++ b/README.md
@@ -348,19 +348,19 @@ TYK_PMP_PUMPS_MONGOAGG_META_USEMIXEDCOLLECTION=true
 ###### Self Healing
 By default, the maximum size of a document in MongoDB is 16MB. If we try to update a document that has grown to this size, an error is received.
 
-The Mongo Aggregate pump creates a new document in the database for each "aggregation period"; the length of that period is defined by `current_aggregation_time`. If, during that period (in minutes) the document grows beyond 16MB, the error will be received and no more records will be recorded until the end of the aggregation period (when a new document will be created).
+The Mongo Aggregate pump creates a new document in the database for each "aggregation period"; the length of that period is defined by `aggregation_time`. If, during that period (in minutes) the document grows beyond 16MB, the error will be received and no more records will be recorded until the end of the aggregation period (when a new document will be created).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
 TYK_PMP_PUMPS_MONGOAGG_META_USEMIXEDCOLLECTION=true
 ```
 
+###### Self Healing
+The default max size of a document in MongoDB is 16Mb. In Mongo Aggregate, when we try to update a document, and the size reaches the limit, an error is received.
```
andyo-tyk, line 349: I think this should be 16MB, not 16Mb.
Is the error something specific to Mongo Aggregate Pump, or just a standard thing with MongoDB? (I'm guessing it's the latter)

I think this line could definitely be re-written to be clearer - and can propose different words if you can confirm the above.

For example, to something like this:

By default, the maximum size of a document in MongoDB is 16MB. If we try to update a document that has grown to this size, an error is received.

README.md
```diff
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
...
+When Self Healing is enabled and a document size hits 16Mb, Pump will create a new document with the current timestamp and reduce the aggregation time configuration by half. So, if the current_aggregation time is equal to 50, it will be reduced to 25. Also, the data that was being written, is sent to the new document created using its timestamp.
```
andyo-tyk, line 350: The Mongo Aggregate pump creates a new document in the database for each "aggregation period"; the length of that period is defined by `current_aggregation_time`. If, during that period (in minutes) the document grows beyond 16MB, the error will be received and no more records will be recorded until the end of the aggregation period (when a new document will be created).

The Self Healing option in the Mongo Aggregate Pump avoids this data loss by monitoring the size of the current document. When this hits the 16MB limit, Pump will automatically create a new document with the current timestamp and start writing to that instead.

When it does this, the Pump will halve the `current_aggregation_time ` so that it aggregates records for half the time period originally configured before creating a new document, reducing the risk of repeatedly hitting the maximum document size.

This self healing is repeatable, however, such that if the document size does reach maximum (16MB) even with the new shorter aggregation period, a new document will be created and the `current_aggregation_time` will be halved again.

The minimum value for `current_aggregation_time` is 1; Self Healing cannot reduce it beyond this value.

README.md
```diff
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
 TYK_PMP_PUMPS_MONGOAGG_META_USEMIXEDCOLLECTION=true
 ```
 
+###### Self Healing
+The default max size of a document in MongoDB is 16Mb. In Mongo Aggregate, when we try to update a document, and the size reaches the limit, an error is received.
+When Self Healing is enabled and a document size hits 16Mb, Pump will create a new document with the current timestamp and reduce the aggregation time configuration by half. So, if the current_aggregation time is equal to 50, it will be reduced to 25. Also, the data that was being written, is sent to the new document created using its timestamp. 
+Now, following this example, new data will be stored into this new document, and a new document will be created every 25 minutes.
```
andyo-tyk, line 351: For example, if the `current_aggregation_time` is configured as 50 (minutes) but the document hits the maximum size (16MB), a new document will be started and the `current_aggregation_time` will be set to 25 (minutes).

README.md
```diff
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
...
+
+Note that if `store_analytics_per_minute` is equal to true, the value of `aggregation_time` will be equal to 1, no matter the value set. This was designed in this way because of backward compatibility.
```
andyo-tyk, line 353: Note that `store_analytics_per_minute` takes precedence over `current_aggregation_time` so if `store_analytics_per_minute` is equal to true, the value of `current_aggregation_time` will be equal to 1 and self healing will not operate.

README.md
mativm02: @andyo-tyk you're right, the error is just a standard limit set by MongoDB.

That new text looks good to me.

tbuchaillot:
@andyo-tyk @marksou can you check this?

andyo-tyk (approved):
Sorry for the delay, I missed any reminder (if there was one) from github. Looks great - thank you.

tbuchaillot:
@mativm02 what's up with this PR?

mativm02:
> @mativm02 what's up with this PR?

I'm waiting for approval. 
I added you as a reviewer, could you take a look at it, please?

tbuchaillot (approved):

PR #513, opened 2022-11-07
[TT-5426] - Improve Tyk Pump demo mode

mativm02:
## Description

Tyk Pump demo mode has been improved. Now, we can configure the number of records per hour, and the number of days. We can enable/disable the TrackPath setting as well.

These are the new flags:
- `demo-track-path`: Boolean flag that disables/enables track path in analytics records. Default: false.
- `demo-days`: Int flag that determines the number of days for the analytics records. Default: 30 days.
- `demo-records-per-hour`: Int flag that determines the number of records per hour. Default: a random number between 300 and 500.

Consider that if the user has "track_all_paths" enabled in their mongo, the `demo-track-path` will be overwritten.  
## Related Issue

https://tyktech.atlassian.net/browse/TT-5426?atlOrigin=eyJpIjoiNjkwYjk3Mjk1Y2RmNDMyMzg1NzBhZmM4ZWNkYTMzNDMiLCJwIjoiaiJ9
## Motivation and Context

It improves the Tyk Pump Demo experience.
## How This Has Been Tested

Unit tests were created. It was also tested by running Tyk Pump Demo using the above flags.
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,13 @@ TYK_PMP_PUMPS_MONGOAGG_TYPE=mongo-pump-aggregate
 TYK_PMP_PUMPS_MONGOAGG_META_USEMIXEDCOLLECTION=true
 ```
 
+###### Self Healing
+The default max size of a document in MongoDB is 16Mb. In Mongo Aggregate, when we try to update a document, and the size reaches the limit, an error is received.
+When Self Healing is enabled and a document size hits 16Mb, Pump will create a new document with the current timestamp and reduce the aggregation time configuration by half. So, if the current_aggregation time is equal to 50, it will be reduced to 25. Also, the data that was being written, is sent to the new document created using its timestamp. 
+Now, following this example, new data will be stored into this new document, and a new document will be created every 25 minutes.
+
+Note that if `store_analytics_per_minute` is equal to true, the value of `aggregation_time` will be equal to 1, no matter the value set. This was designed in this way because of backward compatibility.
+
 ###### Mongo Graph Pump
 As of Pump 1.7+, a new mongo is available called the `mongo_graph` pump. This pump is specifically for parsing 
 GraphQL and UDG requests, tracking information like types requested, fields requested, specific graphql body errors etc.
--- a/analytics/demo/demo.go
+++ b/analytics/demo/demo.go
@@ -149,10 +149,12 @@ func country() string {
 	return codes[rand.Intn(len(codes))]
 }
 
-func GenerateDemoData(start time.Time, days int, recordsPerHour int, orgId string, trackPath bool, writer func([]interface{}, *health.Job, time.Time, int)) {
+func GenerateDemoData(days, recordsPerHour int, orgID string, trackPath bool, writer func([]interface{}, *health.Job, time.Time, int)) {
+	t := time.Now()
+	start := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location())
 	count := 0
-	for d := days; d >= 0; d-- {
-		for h := 0; h < 23; h++ {
+	for d := 0; d < days; d++ {
+		for h := 0; h < 24; h++ {
 			set := []interface{}{}
 			ts := start.AddDate(0, 0, d)
 			ts = ts.Add(time.Duration(h) * time.Hour)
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -350,6 +350,13 @@ TYK_PMP_PUMPS_MONGOAGG_META_AGGREGATIONTIME=50
 TYK_PMP_PUMPS_MONGOAGG_META_ENABLESELFHEALING=true
 ```
 
+###### Self Healing
```
tbuchaillot, line 353: @mativm02 this is not related to this PR

tbuchaillot (changes requested):
@mativm02 good job! it looks good! 
Only one small correction :)

mativm02:
> Does false mean "disable"?

`false` means disable. I added it to the readme.

> Is there a maximum? What happens if this is set to zero?

There is no maximum. If it is set to zero, no records will be created. 

> Is there a maximum or minimum value for this?

There is no maximum or minimum value for this.


As this is a demo, I think users should be free to choose the values they want to use. I mean, if users want to test Tyk Pump generating a really big number of records per hour using their own testing database, they should be able to do it.

Maybe we can set a minimum value of 1 in both `demo-records-per-hour` and `demo-days` in order to always generate records. Another option could be just printing a log saying that `demo-records-per-hour` or `demo-days` is set to 0 and no records will be generated. What do you guys think? @andyo-tyk @tbuchaillot

tbuchaillot:
> > Does false mean "disable"?
> 
> `false` means disable. I added it to the readme.
> 
> > Is there a maximum? What happens if this is set to zero?
> 
> There is no maximum. If it is set to zero, no records will be created.
> 
> > Is there a maximum or minimum value for this?
> 
> There is no maximum or minimum value for this.
> 
> As this is a demo, I think users should be free to choose the values they want to use. I mean, if users want to test Tyk Pump generating a really big number of records per hour using their own testing database, they should be able to do it.
> 
> Maybe we can set a minimum value of 1 in both `demo-records-per-hour` and `demo-days` in order to always generate records. Another option could be just printing a log saying that `demo-records-per-hour` or `demo-days` is set to 0 and no records will be generated. What do you guys think? @andyo-tyk @tbuchaillot

I think the current defaults values are ok - we are not breaking any existing functionality.
I'm +1 on not adding max values, I don't think it's really necessary.

tbuchaillot (approved):

andyo-tyk:
> > As this is a demo, I think users should be free to choose the values they want to use. I mean, if users want to test Tyk Pump generating a really big number of records per hour using their own testing database, they should be able to do it.
> > Maybe we can set a minimum value of 1 in both `demo-records-per-hour` and `demo-days` in order to always generate records. Another option could be just printing a log saying that `demo-records-per-hour` or `demo-days` is set to 0 and no records will be generated. What do you guys think? @andyo-tyk @tbuchaillot
> 
> I think the current defaults values are ok - we are not breaking any existing functionality. I'm +1 on not adding max values, I don't think it's really necessary.

Yep, I've got no problem with not having maximum values - I just wanted to ensure that, if there were, they were documented.

It's a bit complicated, perhaps, that there are so many ways to configure this so that it doesn't create any logs - but as it's just for demo/test usage I think that's fine. It's not a feature anyone will be using (and misconfiguring - and so creating support requests) in production.

Happy to leave it as it is. :)

README.md
```diff
@@ -1231,4 +1231,4 @@ You can run Tyk Pump in demo mode, which will generate fake analytics data and s
 - `--demo-api=<API_ID>` - Sets the API ID to use for the demo data. If not set, a random API ID will be generated.
```
andyo-tyk, line 1231: I think this means that this <API_ID> will be recorded in the demo transactions.
I would re-word this slightly to:
`Configure the value to be recorded as the <API_ID> in the demo transactions. If this option is not set, the Pump Demo mode will use a random <API_ID>. Note that the same <API_ID> will be used for all transaction logs.`

README.md
```diff
@@ -1231,4 +1231,4 @@ You can run Tyk Pump in demo mode, which will generate fake analytics data and s
...
 - `--demo-days=<DAYS>` - Sets the number of days of demo data to generate. Defaults to 30.
 - `--demo-records-per-hour=<RECORDS_PER_HOUR>` - Sets the number of records to generate per hour. The default value is a random number between 300 and 500.
-- `--demo-track-path` - Enables tracking of the request path in the demo data. Defaults to false. Consider that having `track_all_paths` set to true in your pump configuration will override this option.
+- `--demo-track-path` - Enables tracking of the request path in the demo data. Defaults to false (disabled). Consider that having `track_all_paths` set to true in your pump configuration will override this option.
```
andyo-tyk, line 1234: Please would you change the last sentence to:
Note that setting `track_all_paths` to `true` in your Pump configuration will override this option.

andyo-tyk (changes requested):

README.md
mativm02: done!

furkansenharputlu (approved):

andyo-tyk (approved):

PR #519, opened 2022-11-11
[TT-5429] Tyk Pump Ignore Fields 

mativm02:
## Description

We are allowing users to ignore analytics record fields on their pumps. They can set this with any JSON tag of the [analytic record struct](https://github.com/TykTechnologies/tyk-pump/blob/master/analytics/analytics.go#L30) to ignore that field.
## Related Issue

https://tyktech.atlassian.net/jira/software/c/projects/TT/boards/42?modal=detail&selectedIssue=TT-5429
## Motivation and Context

It allows users to ignore fields that they don't really need.
## How This Has Been Tested

Unit tests were created and it was also tested with Pump Demo Mode and Mongo Pump.
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"fmt"
 	"net"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
 
+	"github.com/fatih/structs"
 	"github.com/oschwald/maxminddb-golang"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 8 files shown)


analytics/analytics.go
```diff
@@ -348,3 +349,23 @@ func (a *AnalyticsRecord) IsGraphRecord() bool {
 
 	return false
 }
+
+func (a *AnalyticsRecord) RemoveIgnoredFields(ignoreFields []string) {
```
tbuchaillot, line 353: @mativm02 I think using reflection here makes this change way more dangerous. Can you try to refactor it using `github.com/fatih/structs`? We already use that lib in this type of stuff (for example https://github.com/TykTechnologies/tyk-pump/blob/1f0bc24934cd6d09829f5c0f6d6fe2e243f6b7ac/analytics/aggregate.go#L160), and IMO it's much safer

tbuchaillot (changes requested):
Added some comment

analytics/analytics.go
tbuchaillot: Nope @mativm02 - we didn't have any problem related to that repo (yet - I don't want to jinx it 😅 )

mativm02: @tbuchaillot done! now it's implementing `structs` library

tbuchaillot (approved):

sredxny (approved):
lgtm

PR #520, opened 2022-11-11
TT-6834 Timeout documentation

tbuchaillot:
## Description

Changing the wording and the format/example to a more generic one since this configuration applies to all pumps.

## Related Issue

https://tyktech.atlassian.net/browse/TT-6834
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/config.go
+++ b/config.go
@@ -60,8 +60,8 @@ type PumpConfig struct {
 	//
 	// The format for this configuration is:
 	// ```{.json}
-	// "{pump_name}": {
-	//   "type": "{pump_type}",
+	// "pump_name": {
+	//   ...
 	//   "timeout":5,
 	//   "meta": {...}
 	// }
--- a/config.go
+++ b/config.go
@@ -54,11 +54,10 @@ type PumpConfig struct {
 	// }
 	// ```
 	Filters analytics.AnalyticsFilters `json:"filters"`
-	// You can configure a different timeout for each pump with the configuration option `timeout`.
-	// Its default value is `0` seconds, which means that the pump will wait for the writing
-	// operation forever.
+	// By default, a pump will wait forever for each write operation to complete; you can configure an optional timeout by setting the configuration option timeout.
+	// If you have deployed multiple pumps, then you can configure each timeout independently. The timeout is in seconds and defaults to 0.
 	//
-	// The format for this configuration is:
+	// The timeout is configured within the main pump config as shown here; note that this example would configure a 5 seconds timeout:
 	// ```{.json}
 	// "pump_name": {
 	//   ...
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -58,26 +58,23 @@ type PumpConfig struct {
...
-	//   }
+	//   "meta": {...}
 	// }
 	// ```
 	//
 	// In case that any pump doesn't have a configured timeout, and it takes more seconds to write
 	// than the value configured for the purge loop in the `purge_delay` config option, you will
-	// see the following warning message: `Pump PMP_NAME is taking more time than the value
+	// see the following warning message: `Pump {pump_name} is taking more time than the value
 	// configured of purge_delay. You should try to set a timeout for this pump.`.
```
andyo-tyk, line 73: Perhaps replace this paragraph with:

Tyk will inform you if the pump's write operation is taking longer than the purging loop (configured via `purge_delay`) as this will mean that data is purged before being written to the target data sink.

If there is no `timeout` configured, the following warning log will be generated if this occurs:
`Pump {pump_name} is taking more time than the value configured of purge_delay. You should try to set a timeout for this pump.`

andyo-tyk (changes requested):
Would you please consider these tweaks to the wording in this section?

config.go
```diff
@@ -58,26 +58,23 @@ type PumpConfig struct {
...
 	//
 	// In case that you have a configured timeout, but it still takes more seconds to write than
 	// the value configured for the purge loop in the `purge_delay` config option, you will see the
-	// following warning message: `Pump PMP_NAME is taking more time than the value configured of
+	// following warning message: `Pump {pump_name} is taking more time than the value configured of
 	// purge_delay. You should try lowering the timeout configured for this pump.`.
```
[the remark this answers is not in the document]
tbuchaillot: got it! solved ✅

config.go
```diff
@@ -54,11 +54,10 @@ type PumpConfig struct {
 	// }
 	// ```
 	Filters analytics.AnalyticsFilters `json:"filters"`
-	// You can configure a different timeout for each pump with the configuration option `timeout`.
-	// Its default value is `0` seconds, which means that the pump will wait for the writing
-	// operation forever.
+	// By default, a pump will wait forever for each write operation to complete; you can configure an optional timeout by setting the configuration option timeout.
```
andyo-tyk, line 57: Could you please put the last _timeout_ in `code case`, as it's the name of the key?

config.go
```diff
@@ -54,11 +54,10 @@ type PumpConfig struct {
...
+	// If you have deployed multiple pumps, then you can configure each timeout independently. The timeout is in seconds and defaults to 0.
 	//
-	// The format for this configuration is:
+	// The timeout is configured within the main pump config as shown here; note that this example would configure a 5 seconds timeout:
```
andyo-tyk, line 60: It's another subtle quirk of English: this should be _5 second timeout_ (singular), not _5 seconds timeout_ (plural).

andyo-tyk (changes requested):

andyo-tyk (commented):
This bit is now perfect - thank you.

config.go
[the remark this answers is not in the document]
andyo-tyk: Perfect - thank you.

andyo-tyk (approved):
Thanks

marksou (approved):
LGTM

PR #558, opened 2023-01-04
TT-7216 Decode Option For Raw Request/Response

Firaasss:
## Description
Added an option to decode the raw request and raw response data of each analytical record from base64 prior to writing them to the data sink. This solution is implemented at the same level as other general pump filters. 

A configuration option has been included to allow the decoding feature to be turned on or off (off by default) in `config.go`.  

## Related Issue
https://tyktech.atlassian.net/browse/TT-7216

## Motivation and Context
This gives additional functionality to our Tyk-Pump component to allow the encoded transactional logs to be decoded for human readability. This ticket covers two previous requests (from community and customers) that were specifically for syslog and csv. Both of these reference the fact that this functionality is already implemented for the Elasticsearch pump.

## How This Has Been Tested
There are a variety of tests cases that are made in `main_test.go` to generate scenarios that will test decoding the raw request/response in 4 different scenarios. 

Another test using an stdout pump while leveraging the pump demo (using --demo tag) was done to test if the values from the config are actually working. This was done using a redis and mongo service to verify the values that were being written after setting the config value and decoding the raw response/request. 

Eg. Setting `"raw_request_decoded": false` and `"raw_response_decoded": true` in `pump.conf`

Screen Shot 2023-01-05 at 1 15 05 PM

## Types of changes

## Checklist

  - [x] If you've changed APIs, describe what needs to be updated in the documentation.
  - [ ] `go fmt -s`
  - [ ] `go vet`


main.go
```diff
@@ -300,7 +302,10 @@ func filterData(pump pumps.Pump, keys []interface{}) []interface{} {
 	shouldTrim := SystemConfig.MaxRecordSize != 0 || pump.GetMaxRecordSize() != 0
 	filters := pump.GetFilters()
 	ignoreFields := pump.GetIgnoreFields()
-	if !filters.HasFilter() && !pump.GetOmitDetailedRecording() && !shouldTrim && len(ignoreFields) == 0 {
+	getDecodingResponse := pump.GetDecodedResponse()
+	getDecodingRequest := pump.GetDecodedRequest()
+	if !getDecodingRequest && !getDecodingResponse && !filters.HasFilter() && !pump.GetOmitDetailedRecording() && !shouldTrim && len(ignoreFields) == 0 {
+		fmt.Println("Returning here")
```
sredxny, line 308: please remove this debug line

sredxny (changes requested):
lgtm but requested a small change

mativm02 (approved):
well done @Firaasss! please, remove that testing print :) overall, an amazing job!

main.go
```diff
@@ -300,7 +302,10 @@ func filterData(pump pumps.Pump, keys []interface{}) []interface{} {
 	shouldTrim := SystemConfig.MaxRecordSize != 0 || pump.GetMaxRecordSize() != 0
 	filters := pump.GetFilters()
 	ignoreFields := pump.GetIgnoreFields()
-	if !filters.HasFilter() && !pump.GetOmitDetailedRecording() && !shouldTrim && len(ignoreFields) == 0 {
+	getDecodingResponse := pump.GetDecodedResponse()
+	getDecodingRequest := pump.GetDecodedRequest()
+	if !getDecodingRequest && !getDecodingResponse && !filters.HasFilter() && !pump.GetOmitDetailedRecording() && !shouldTrim && len(ignoreFields) == 0 {
```
sredxny, line 307: also, could you explain the logic in this conditional? it can be an inline comment or something, as it's a pretty long conditional

sredxny (dismissed):

main.go
Firaasss: Removed and pushed.

main.go
Firaasss: Yes done - my addition to this code was `!getDecodingRequest `and `!getDecodingResponse` to check if the configs for both of those was false. Essentially the whole conditional is to see if all config options are empty or false.

PR #563, opened 2023-01-16
Increasing test coverage of logger package

mativm02:
## Description

I've created a unit test for GetLogger() function. I've also deleted some unnecessary lines of code. Whenever TYK_LOGLEVEL has an invalid value (empty or a not valid string), the log level will be set as "info". That's why we don't need to add an "info" case in the switch sentence, we can just keep it as default.

**Before:**
212693397-5371fa31-f0c5-4df0-bbbb-eb74eb4f7cd2

**After:**
212693338-47593ddc-8ac9-4481-acf9-69220a804747

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/logger/init.go
+++ b/logger/init.go
@@ -24,17 +24,12 @@ func GetFormatterWithForcedPrefix() *prefixed.TextFormatter {
 
 func GetLogger() *logrus.Logger {
 	level := os.Getenv("TYK_LOGLEVEL")
-	if level == "" {
-		level = "info"
-	}
 
 	switch strings.ToLower(level) {
 	case "error":
 		log.Level = logrus.ErrorLevel
 	case "warn":
 		log.Level = logrus.WarnLevel
-	case "info":
-		log.Level = logrus.InfoLevel
 	case "debug":
 		log.Level = logrus.DebugLevel
 	default:
--- a/logger/init_test.go
+++ b/logger/init_test.go
@@ -88,6 +88,7 @@ func Test_GetLooger(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			os.Setenv("TYK_LOGLEVEL", tt.env)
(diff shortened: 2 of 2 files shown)


logger/init.go
```diff
@@ -24,17 +24,12 @@ func GetFormatterWithForcedPrefix() *prefixed.TextFormatter {
 
 func GetLogger() *logrus.Logger {
 	level := os.Getenv("TYK_LOGLEVEL")
-	if level == "" {
```
tbuchaillot, line 27: @mativm02 why is this removed?

mativm02: Because "info" is the default value. If the user sends an empty string, we don't need to convert it to "info", since it will join the "default" case in the switch sentence, which set the log level to "InfoLevel".
That's also the reason why I've removed `case "info":` in the switch sentence.

logger/init_test.go
```diff
@@ -51,3 +51,47 @@ func TestFormatterWithForcedPrefixFileOutput(t *testing.T) {
...
+		{
+			name:          "debug",
+			env:           "debug",
+			expectedLevel: logrus.DebugLevel,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			os.Setenv("TYK_LOGLEVEL", tt.env)
```
tbuchaillot, line 90: @mativm02 missed the defer of unset of the env after the test

logger/init.go
tbuchaillot: makes sense!

tbuchaillot (changes requested):

logger/init_test.go
mativm02: @tbuchaillot I will add it! thanks

tbuchaillot (approved):

sredxny (approved):
lgtm

PR #588, opened 2023-03-15
TT-8313 Hybrid pump refactor

tbuchaillot:
## Description

This PR drops the `gateway` dependency for `Hybrid Pump`, now it uses directly `gorpc` lib for MDCB connection.

It cleans a little bit Hybrid pump code:
 - Parse the config as the rest of the pumps. Now it's going to be possible to auto-generate docs for the Hybrid pump.
 - It tries `Login` on pump Init.  No reason to maintain the pump if the credentials are incorrect.
- It always connects using the `Login` function.
- Drops the `Ping` function call on each WriteData => we are already checking connection with `Login`. It just checks if `Ping` results in error when connecting.
- Adds retries mechanism on pump `Init` with exponential backoff and 3 retries.

Adds tests using a mockedRPC Server:
- `TestHybridPumpInit` -> checks pump basic Init behavior
- `TestConfigParsing` ->  checks env vars / config setting
- `TestHybridPumpWriteData` -> checks aggregate/non-aggregate data writing
- `TestHybridPumpShutdown` -> checks connection closing on shutdown
- `TestWriteLicenseExpire` -> checks what happens if the RPC `Login` starts to say that the node is non-authorized.

## Related Issue

https://tyktech.atlassian.net/browse/TT-8313
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/hybrid_test.go
```diff
@@ -0,0 +1,424 @@
...
+	"fmt"
+	"io"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/TykTechnologies/gorpc"
+	"github.com/TykTechnologies/tyk-pump/analytics"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/atomic"
```
titpetric, line 15: Test only dep, stdlib is better

titpetric (approved):

go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/TykTechnologies/tyk-pump
 
-go 1.16
+go 1.15
```
buger, line 3: Why Go downgrade?

buger:
Thus RPC layer works using serialize which depends on native stucts, and if they do not match will throw error. 
So will this hybrid pump will work with older mdcb/gateway?

go.mod
tbuchaillot: whoops

tbuchaillot:
> Thus RPC layer works using serialize which depends on native stucts, and if they do not match will throw error. 

@buger, we are not sending any structure in the RPC layer here -
`Login` receives credentials as [string](https://github.com/TykTechnologies/tyk-sink/blob/master/dispatcher/dispatcher.go#L40). 
`PurgeAnalyticsData` and `PurgeAnalyticsDataAggregated ` receive the analytic data marshaled as a [string](https://github.com/TykTechnologies/tyk-sink/blob/master/dispatcher/dispatcher.go#LL122C20-L122C48). 



> So will this hybrid pump will work with older mdcb/gateway?

it should 😅 tested on the current MDCB/Gw without any change + MDCB 1.8.1/Gw3.2.3 and it's working.

 It's not modifying the content of what we sent before - the only thing we are dropping is the `LoginWithGroup` since it wasn't even working before, and it makes more sense to do the `Login` here.

pumps/hybrid.go
```diff
@@ -39,18 +35,71 @@ var (
...
+	if conf.CallTimeout == 0 {
+		conf.CallTimeout = GlobalRPCCallTimeout
+	}
+
+	if conf.Aggregated {
+		if conf.StoreAnalyticsPerMinute {
+			conf.aggregationTime = 1
+		} else {
+			conf.aggregationTime = 60
+		}
```
sredxny, line 101: it could be rewriten it as:

```
conf.aggregationTime = 60
if conf.StoreAnalyticsPerMinute {
	conf.aggregationTime = 1
}
```

pumps/hybrid.go
```diff
@@ -62,121 +111,145 @@ func (p *HybridPump) New() Pump {
...
+	if errConnect != nil {
+		p.log.Fatal("Failed to connect to RPC server")
 	}
-	if groupID, ok := meta["group_id"]; ok {
-		rpcConfig.GroupID = groupID.(string)
+	p.log.Info("starting rpc dispatcher")
+	p.startDispatcher()
+
+	p.log.Info("loging in to MDCB rpc server")
+	logged, err := p.RPCLogin()
```
sredxny, line 142: what would happen if the connection get lost after this? what if we continue checking that we can Login to the server, and if it fails then we attempt to reconnect?

sredxny (changes requested):

pumps/hybrid.go
```diff
@@ -39,18 +35,71 @@ var (
...
+
+	// Specifies prefixes of tags that should be ignored if `aggregated` is set to `true`.
+	IgnoreTagPrefixList []string `json:"ignore_tag_prefix_list" mapstructure:"ignore_tag_prefix_list"`
+
+	// RPC server call timeout
+	CallTimeout int `mapstructure:"call_timeout"`
+	// RPC server connection pool size
+	RPCPoolSize int `mapstructure:"rpc_pool_size"`
+
+	aggregationTime int
```
mativm02, line 74: can we add a comment on this field? (despite it's unexported)

mativm02 (changes requested):

pumps/hybrid.go
titpetric: Suggested to make it a func in another place; the value hides logic.

tbuchaillot: added comment

pumps/hybrid.go
```diff
@@ -39,18 +35,71 @@ var (
 			return nil
 		},
 	}
+	GlobalRPCCallTimeout = 30
```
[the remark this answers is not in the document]
tbuchaillot: renamed and defaulted to 10sec

pumps/hybrid.go
tbuchaillot: done!

pumps/hybrid.go
```diff
@@ -230,3 +305,26 @@ func (p *HybridPump) WriteData(ctx context.Context, data []interface{}) error {
 
 	return nil
 }
+
+func (p *HybridPump) Shutdown() error {
```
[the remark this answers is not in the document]
tbuchaillot: Shutdown is is executed at the main level for all the pumps https://github.com/TykTechnologies/tyk-pump/blob/master/main.go#L259, aka when we are gracefully stopping Tyk Pump as a whole.

sredxny (approved):
🚀

mativm02 (approved):
lgtm!

PR #651, opened 2023-05-31
[TT-8999] - Converting pump name to lower case when looking for it

mativm02:
## Description

We have a map `AvailablePumps` that contains all the available pumps. The key of the map is the type written in lowercase (mongo, csv, prometheus, etc.) and the value is an instance of the pump itself:
https://github.com/TykTechnologies/tyk-pump/blob/54d0455505f8c864e9742ce608c9428560e6a101/pumps/init.go#L9

Since [this PR](https://github.com/TykTechnologies/tyk-pump/pull/648), we're defaulting all the pumps to uppercase in order to avoid repeated values and to allow users to override pumps through env vars. That's why we need to temporally make sure that the type is lowercase in order to get the struct from the AvailablePumps map.
## Related Issue

https://tyktech.atlassian.net/browse/TT-8999?atlOrigin=eyJpIjoiMWQ2NDc5MmY3YzdjNGU5MDkzNWIwOWNmNDQzYmU5ZjgiLCJwIjoiaiJ9
## How This Has Been Tested

Tested using Pump with and without env vars, with different config files.
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/pump.go
+++ b/pumps/pump.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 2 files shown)


mativm02:
> can we add a test?

@sredxny added!

sredxny (approved):

PR #694, opened 2023-07-27
[TT-9468] New SQL Aggregate indexes

tbuchaillot:
## Description

This PR drops the `dimension` index for sql_aggregate pumps and creates a new one called `idx_dimension`. This change is made so the index we are using has the same order as Tyk Dashboard queries.

In order to avoid blocking production table, we execute the index creation in a goroutine when the database is `postgres` and use the special postgres instruction `CONCURRENTLY` to avoid table locking. More info about this [here](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)

## Related Issue

https://tyktech.atlassian.net/browse/TT-9468
## Motivation and Context

https://tyktech.atlassian.net/browse/TT-9468
## How This Has Been Tested

Added unit tests for SQLite scenarios (we don't have Postgres running on CI)
Tested manually

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


sredxny (approved):

pumps/sql_aggregate.go
```diff
@@ -119,6 +145,64 @@ func (c *SQLAggregatePump) Init(conf interface{}) error {
...
+// if not, it will block until it finishes
+func (c *SQLAggregatePump) ensureIndex(tableName string, background bool) error {
+	if !c.db.Migrator().HasIndex(tableName, newAggregatedIndexName) {
+		createIndexFn := func(c *SQLAggregatePump) error {
+			option := ""
+			if c.dbType == "postgres" {
+				option = "CONCURRENTLY"
+			}
+
+			err := c.db.Table(tableName).Exec(fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS %s ON %s (dimension, timestamp, org_id, dimension_value)", option, newAggregatedIndexName, tableName)).Error
```
mativm02, line 160: if `c.dbType` is different than "postgres", `option` would be an empty string and this will probably fail (didn't reproduce it) - is that expected?

pumps/sql_aggregate_test.go
```diff
@@ -326,3 +335,105 @@ func TestDecodeRequestAndDecodeResponseSQLAggregate(t *testing.T) {
 	assert.False(t, newPump.GetDecodedRequest())
 	assert.False(t, newPump.GetDecodedResponse())
 }
+
+func TestEnsureIndex(t *testing.T) {
```
mativm02, line 339: is it possible to add any failure scenario? If not, remove the `expectedErr` field

mativm02 (changes requested):

pumps/sql_aggregate.go
tbuchaillot: It doesn't fail since `CREATE INDEX IF NOT EXISTS ` is a valid query for other dbTypes.

tbuchaillot:
@CodiumAI-Agent review

pumps/sql_aggregate_test.go
tbuchaillot: added

.github/workflows/linter.yaml
```diff
@@ -27,14 +27,13 @@ jobs:
       go: ${{ matrix.go }}
       redis: 5
       mongo: 4.2
-      test-options: ""-count=1 -failfast -v""
```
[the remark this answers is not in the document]
tbuchaillot: reverted

mativm02 (approved):

mativm02 (approved):

sredxny (approved):
lgtm

PR #696, opened 2023-07-27
[TT-9360] Changing Timeout from time.Duration to int

mativm02:
## Description

Changing Timeout from time.Duration to int
## Related Issue

https://tyktech.atlassian.net/jira/software/c/projects/TT/boards/42?modal=detail&selectedIssue=TT-9360
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


pumps/kafka.go
```diff
@@ -137,9 +142,21 @@ func (k *KafkaPump) Init(config interface{}) error {
 		k.log.WithField("SASL-Mechanism", k.kafkaConf.SASLMechanism).Warn("Tyk pump doesn't support this SASL mechanism.")
 	}
 
+	// Timeout is an interface type to allow both time.Duration and float values
```
mativm02, line 145: should we add a default value here? there wasn't a default one before this change

pumps/kafka.go
```diff
@@ -90,6 +91,10 @@ func (k *KafkaPump) Init(config interface{}) error {
 	}
 
 	processPumpEnvVars(k, k.log, k.kafkaConf, kafkaDefaultENV)
+	// This interface field is not reached by envconfig library, that's why we manually check it
+	if os.Getenv("TYK_PMP_PUMPS_KAFKA_META_TIMEOUT") != "" {
```
mativm02, line 95: I'm not too fond of this way of getting the env var, but `envconfig` library isn't getting the env var if the data type is an `interface`

pumps/kafka.go
```diff
@@ -137,9 +142,21 @@ func (k *KafkaPump) Init(config interface{}) error {
...
+	switch k.kafkaConf.Timeout.(type) {
+	case string:
+		timeout, err = time.ParseDuration(k.kafkaConf.Timeout.(string))
+		if err != nil {
+			k.log.Fatal("Failed to parse timeout: ", err)
+		}
+	case float64:
+		timeout = time.Duration(k.kafkaConf.Timeout.(float64)) * time.Second
+	}
+
```
sredxny, line 156: if not string or float, it will default to `0s` is that expected?

sredxny (changes requested):
check tests failing

pumps/kafka.go
mativm02: @sredxny it was the previous behavior before this change. Just in case, `0s` means no timeout from the Kafka library documentation

sredxny (approved):

PR #743, opened 2023-11-15
[SYSE-294] Standardise build time variables

alephnull:
These variables are set at build time. These changes help the releng templates use a standard build argument. See https://tyktech.slack.com/archives/C03PT641CF9/p1699582096160529


ermirizio (approved):
Smoke tests fixed, linter issue fixed (thanks @mativm02 ). LGTM

PR #745, opened 2023-11-20
TT-10564 Splunk backoff retry

domsolutions:
Introduce an expotential backoff retry mechanism for sending logs to spunk to mitigate the chance of losing logs if pump fails to send logs to spunk (i.e. due to network issue or rate limit being hit). Also fixes bug where pump not closing connections after receiving response. Also spunk responses were not being checked i.e. a `404` could be returned and pump would still mark as successful pump export.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


tbuchaillot:
hey @domsolutions thanks for the contribution! we will have a look and come back to you :)

domsolutions:
Hi @tbuchaillot any ETA when this can be looked at?

tbuchaillot:
> Hi @tbuchaillot any ETA when this can be looked at?

Hey @domsolutions , we are starting a new sprint today and the review/QA of this will be included.

domsolutions:
Hi @tbuchaillot sorry to chase but we're quite eager for this fix, when does your sprint end so I can have an idea when it'll be looked at?

mativm02:
Hi @domsolutions! I've been taking a look at the code and it looks pretty good overall!

I just found a bug related to the retry logic in the `Send` function. When the function retries a request after the first attempt, it encounters an error with the message "http: ContentLength=444 with Body length 0". This happens because the body of an HTTP request in Go is read as an `io.Reader`, which can only be read once. After the first read, the body is exhausted and can't be read again for subsequent retries, leading to a mismatch between the expected content length and the actual empty body.

To address this, we would need to ensure that the request body is reset or recreated for each retry. One approach would be to buffer the entire body before sending the initial request, and then use this buffer to create a new `io.Reader` for each retry. This would ensure that the full body content is available for each attempt, thus avoiding the mismatch error.

I'm thinking on something like this:
```
func (s *BackoffHTTPRetry) Send(req *http.Request) error {
	// Read and buffer the entire request body
	var requestBodyBytes []byte
	if req.Body != nil {
		requestBodyBytes, _ = io.ReadAll(req.Body)
		req.Body.Close() // closing the original body
	}
	opFn := func() error {
		// recreating the request body from the buffer for each retry
		req.Body = io.NopCloser(bytes.NewBuffer(requestBodyBytes))
		resp, err := s.httpclient.Do(req)
		if err != nil {
			return s.handleErr(err)
		}
		defer func() {
			// read all response and discard so http client can
			// reuse connection as per doc on Response.Body
			io.Copy(io.Discard, resp.Body)
			resp.Body.Close()
		}()
                
                ...
```
If you have any questions or need further clarification, feel free to ask!

Thanks for your contribution, and I'm looking forward to the revised implementation!

domsolutions:
Hi @mativm02 

Hmm that's a strange one as I'm not able to replicate. I understand the issue but the core Go library deals with this issue where it resets the body https://github.com/golang/go/blob/go1.19/src/net/http/transport.go#L621C14-L621C24 which version of Go are you using? I'll download and try and replicate.

domsolutions:
> Hi @mativm02
> 
> Hmm that's a strange one as I'm not able to replicate. I understand the issue but the core Go library deals with this issue where it resets the body https://github.com/golang/go/blob/go1.19/src/net/http/transport.go#L621C14-L621C24 which version of Go are you using? I'll download and try and replicate.

Also what was the test scenario to see that issue?

mativm02:
> Hi @mativm02
> 
> Hmm that's a strange one as I'm not able to replicate. I understand the issue but the core Go library deals with this issue where it resets the body https://github.com/golang/go/blob/go1.19/src/net/http/transport.go#L621C14-L621C24 which version of Go are you using? I'll download and try and replicate.

@domsolutions I tried using Go v1.19.10 and 1.20.5 - what I'm doing is:
1. I ran a Tyk Gateway instance
2. I created an API with the local Splunk endpoint as upstream
3. I set a rate limit to the API so I can get 429 HTTP statuses.
4. Then I ran tyk-pump in demo mode (`go build && ./tyk-pump --conf ../tyk-develop-env/confs/pump_splunk.conf --demo=63cec3ef95014f71aa7d48b9 --demo-api=64888c2809b5b0000101f29a --demo-records-per-hour=1 --demo-days=1`) - it should work on your end using the same command.
5. On the second iteration, the retry system fails with this error:
`time="Dec 27 08:55:23" level=warning msg="Error Writing to: Splunk Pump - Error:Post \"https://localhost:8181/splunk/\": http: ContentLength=442 with Body length 0" prefix=main` and it stops retrying (although I set `max_retries` to 3)

Adding the buffer solved the issue - setting `req.ContentLength = 0` also did it (that's not a solution, just a workaround).

domsolutions:
> > Hi @mativm02
> > Hmm that's a strange one as I'm not able to replicate. I understand the issue but the core Go library deals with this issue where it resets the body https://github.com/golang/go/blob/go1.19/src/net/http/transport.go#L621C14-L621C24 which version of Go are you using? I'll download and try and replicate.
> 
> @domsolutions I tried using Go v1.19.10 and 1.20.5 - what I'm doing is:
> 
> 1. I ran a Tyk Gateway instance
> 2. I created an API with the local Splunk endpoint as upstream
> 3. I set a rate limit to the API so I can get 429 HTTP statuses.
> 4. Then I ran tyk-pump in demo mode (`go build && ./tyk-pump --conf ../tyk-develop-env/confs/pump_splunk.conf --demo=63cec3ef95014f71aa7d48b9 --demo-api=64888c2809b5b0000101f29a --demo-records-per-hour=1 --demo-days=1`) - it should work on your end using the same command.
> 5. On the second iteration, the retry system fails with this error:
>    `time="Dec 27 08:55:23" level=warning msg="Error Writing to: Splunk Pump - Error:Post \"https://localhost:8181/splunk/\": http: ContentLength=442 with Body length 0" prefix=main` and it stops retrying (although I set `max_retries` to 3)
> 
> Adding the buffer solved the issue - setting `req.ContentLength = 0` also did it (that's not a solution, just a workaround).

@mativm02 I was able to replicate in the unit tests by setting on the HTTP server

```golang
server.Config.SetKeepAlivesEnabled(false)
```

So when the server closes the connection after first failed attempt, the client creates a new connection, because it's not resued it doesn't rewind the body https://github.com/golang/go/blob/go1.19.13/src/net/http/transport.go#L699 not sure if that's intentional or a bug in Go but I implemented your suggestion and it fixes it. 

I also added some more rules for retrying so less "retryable" errors aren't missed.

domsolutions:
@mativm02 Does it work for you now?

http-retry/http-retry.go
```diff
@@ -0,0 +1,133 @@
...
+// NewBackoffRetry Creates an exponential backoff retry to use httpClient for connections. Will retry if a temporary error or
+// 5xx or 429 status code in response.
+func NewBackoffRetry(errMsg string, maxRetries uint64, httpClient *http.Client, logger *logrus.Entry) *BackoffHTTPRetry {
+	return &BackoffHTTPRetry{errMsg: errMsg, maxRetries: maxRetries, httpclient: httpClient, logger: logger}
+}
+
+func (s *BackoffHTTPRetry) Send(req *http.Request) error {
+	var reqBody []byte
+	if req.Body != nil {
+		reqBody, _ = io.ReadAll(req.Body)
```
mativm02, line 40: check possible error here to make it safer (and to avoid linting issues)

http-retry/http-retry.go
```diff
@@ -0,0 +1,133 @@
...
+			// reuse connection as per doc on Response.Body
+			io.Copy(io.Discard, resp.Body)
+			resp.Body.Close()
+		}()
+
+		if resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		body, _ := io.ReadAll(resp.Body)
```
mativm02, line 65: check possible error here to make it safer (and to avoid linting issues)

http-retry/http-retry.go
```diff
@@ -0,0 +1,133 @@
...
+		req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
+
+		resp, err := s.httpclient.Do(req)
+		if err != nil {
+			return s.handleErr(err)
+		}
+		defer func() {
+			// read all response and discard so http client can
+			// reuse connection as per doc on Response.Body
+			io.Copy(io.Discard, resp.Body)
```
mativm02, line 57: check possible error here to make it safer (and to avoid linting issues)

pumps/splunk_test.go
```diff
@@ -48,6 +51,14 @@ func (h *testHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		h.test.Fatal("Couldn't ready body")
 	}
+	r.Body.Close()
+
+	if h.returnErrorOnFirstReq && h.reqCount == 1 {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("splunk internal error"))
```
mativm02, line 58: check possible error here to make it safer (and to avoid linting issues)

http-retry/http-retry.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/cenkalti/backoff/v4"
+	"github.com/sirupsen/logrus"
+)
+
+type BackoffHTTPRetry struct {
+	errMsg     string
+	maxRetries uint64
+	logger     *logrus.Entry
+	httpclient *http.Client
+}
```
mativm02, line 23: modify this struct to:
```
type BackoffHTTPRetry struct {
	httpclient *http.Client
	logger     *logrus.Entry
	errMsg     string
	maxRetries uint64
}
```

all pointers should come first to optimize GC

mativm02 (changes requested):

mativm02:
> @mativm02 Does it work for you now?

It works on my end now. Please, address the suggested changes and feel free to ask me for a new review. I'll be happy to approve it. Thank you for your time and work on this, it's really appreciated!

domsolutions:
> > @mativm02 Does it work for you now?
> 
> @domsolutions It works on my end now. Please, address the suggested changes and feel free to ask me for a new review. I'll be happy to approve it. Thank you for your time and work on this, it's really appreciated!

No problem @mativm02 was a pleasure, I've made your requested changes, please take a look when you have a chance, cheers.

mativm02 (approved):
LGTM, thank you @domsolutions! 🚀

domsolutions:
> LGTM, thank you @domsolutions! 🚀

@mativm02 I had to fix a golinter issue to fix the pipeline, does it need approving again?

mativm02:
@domsolutions I created a duplicated version of this PR with all your commits so CIs can effectively run without complaining about some repo secrets. I'll close this one so we can keep working on the new one and merge it. Again, thank you!

https://github.com/TykTechnologies/tyk-pump/pull/758

PR #757, opened 2023-12-21
[SYSE-87] test-controller job decides test scope

alephnull:
The complete list of features for _tyk-pump:master_: releng.

:robot: Beep, boop, this PR has been generated by [automation](https://github.com/TykTechnologies/gromit/blob/master/cmd/policy.go). Changes should be limited to templated code. Where allowed, this PR will be automatically merged when the branch protection rules are satisfied. If you think this was made by mistake, or if merging this might cause any instabilities, please bring it up on Slack or log a jira in the TD project.


buger:
**API tests result - mongo44-sha256 env: skipped** :no_entry_sign: 
Branch used: refs/pull/757/merge
Commit: 7f037014a93f9f3a3721032cff3e9ca518aefc0f 
Triggered by: pull_request (@alephnull)
[Execution page](https://github.com/TykTechnologies/tyk-pump/actions/runs/7292368818)

buger:
**API tests result - postgres15-sha256 env: skipped** :no_entry_sign: 
Branch used: refs/pull/757/merge
Commit: 7f037014a93f9f3a3721032cff3e9ca518aefc0f 
Triggered by: pull_request (@alephnull)
[Execution page](https://github.com/TykTechnologies/tyk-pump/actions/runs/7292368818)

.github/workflows/release.yml
```diff
@@ -173,20 +173,97 @@ jobs:
...
+      - name: set version
+        id: versions
+        env:
+          GW_OR_DASH: ${{ (endsWith(github.repository, 'tyk') || endsWith(github.repository, 'tyk-analytics')) && 'yes' }}
+          BASE_REF: ${{ startsWith(github.event_name, 'pull_request') && github.base_ref || github.ref }}
+          TAGS: ${{ needs.goreleaser.outputs.tags }}
+          # Cover pull_request_target too
+          PR: ${{startsWith(github.event_name, 'pull_request') && 'yes' }}
+          LTS_PR: ${{(startsWith(github.event_name, 'pull_request') && endsWith(github.base_ref, '-lts')) && 'yes' }}
+          TAG: ${{startsWith(github.ref, 'refs/tags') && 'yes' }}
```
ermirizio, line 195: @alephnull I can't figure out where this vars are being used. Should we removed or is the logic not finished?

.github/workflows/release.yml
```diff
@@ -260,15 +327,15 @@ jobs:
         working-directory: auto
         id: test_execution
         run: |
-          echo "## API tests ${{ matrix.db }} ${{ matrix.conf }}" >> $GITHUB_STEP_SUMMARY
+          echo "### API tests ${{ matrix.db }} ${{ matrix.conf }}" >> $GITHUB_STEP_SUMMARY
           if docker run --rm --network auto_default --env-file pytest.env -v ${{ github.workspace }}/reports:/app/reports \
-             ${{ steps.ecr.outputs.registry }}/tyk-automated-tests:${{ startsWith(github.ref_name, 'release-') && github.ref_name || 'master' }} \
-             pytest -c pytest_ci.ini --ci -m "not local and not dind ${{ matrix.markers }}" > tests.out; then
+             ${{ steps.ecr.outputs.registry }}/tyk-automated-tests:${{ needs.test-controller.outputs.gd_tag }} \
```
ermirizio, line 332: @alephnull I'm afraid this won't work when merging against release code freeze branches like release-4.2 since there's no release-4.2 for tyk-automated-tests, it will just work for LTS branches since that's the only case where release lifecycled is shared between gw, dash and TAT

ermirizio (changes requested):

.github/workflows/release.yml
alephnull: TAGS is not used as we do not (yet) have the test matrix that should be run for tags. The logic is complete.

.github/workflows/release.yml
alephnull: Yes, it will not work for `release-4.2`. https://github.com/TykTechnologies/gromit/pull/242/commits/cc01cc1339270775a19223a9709bbb517457e59f ensures that this never runs for branches like `release-4.2`. AFAIK, these branches are temporary and have no long term need as each non-LTS release supersedes the previous one.

alephnull: Thinking about it further, I have reverted https://github.com/TykTechnologies/gromit/pull/242/commits/cc01cc1339270775a19223a9709bbb517457e59f as this change would prevent the workflow from running on code-freeze branches.
So to address branches like release-4.2, we could address it by tagging a release-4.2 for tyk-automated-tests when the branch is created. I think this is a bad process to follow and the bad practice of creating branches like `release-4.2` which have no value (afaict) should be discouraged.

alephnull: Pushed an implementation based on on regexps. LTS policy says that any minor could be an LTS but at present only minor version 0 is an LTS. If another minor version is declared LTS, then we'll have to use something like `(0|3)` which is not perfect but is good enough, I think.

ermirizio:
@alephnull we need to address the string or required tests, I will do a temporary change to see if $ECR causes any issues over terraform and will let you know asap

image

asutosh (approved):

.github/workflows/release.yml
```diff
@@ -176,17 +156,41 @@ jobs:
...
+    steps:
+      - name: set params
+        id: params
+        env:
+          REPO: ${{ github.repository }}
+          # Cover pull_request_target too
+          BASE_REF: ${{startsWith(github.event_name, 'pull_request') && github.base_ref || github.ref}}
+          TAGS: ${{ needs.goreleaser.outputs.tags }}
+          IS_PR: ${{startsWith(github.event_name, 'pull_request') && 'yes' }}
+          IS_LTSPR: ${{(startsWith(github.event_name, 'pull_request') && endsWith(github.base_ref, '-lts')) && 'yes' }}
```
ermirizio, line 179: @alephnull I think this should be just `IS_LTS` I don't see any IS_LTSPR env on controller.go

ermirizio (changes requested):

.github/workflows/release.yml
alephnull: I think IS_LTSPR makes most sense. controller.go can catch up.

PR #809, opened 2024-04-08
[SYSE-336 master] Followup from March template application

alephnull:
Learning from the application:
Applying templates across all branches and repos runs up against the usage limits. Larger runners are available for Team plans. This needs to be explored.
A local run of release.yml is needed to test updates before pushing it to all repos and waiting for >30mins to find out if it works.
For regular updates which collect disparate code changes, the git commit log in the repos should reflect a summary of changes that went into it.
- [x] Mount go package and build cache into build container
- [x] nektos/act for local runs
- [x] dependabot alerts @devops for only .g/w/release.yml
- [x] r5-lts to test against pump and sink master
- [x] Larger runners for Actions
- [x] pk for byol build
- [x] Converge UI and API tests compose environments in tyk-ci

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,22 +88,22 @@ jobs:
           docker run --rm --privileged -e GITHUB_TOKEN=${{ github.token }} \
           -e GOPRIVATE=github.com/TykTechnologies                                \
           -e GO111MODULE=on \
-          -e DEBVERS='${{ matrix.debvers }}' 	      	      	         \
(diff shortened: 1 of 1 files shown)


ci/Dockerfile.std
```diff
@@ -1,4 +1,5 @@
 # Generated by: gromit policy
+# Generated on: Wed Apr  3 06:48:24 UTC 2024
```
asutosh, line 2: This looks like it is adding the timestamp line, we've ditched the timestamp altogether, right?

asutosh (approved):

Issue #816, opened 2024-05-01
TT-12057 Splunk pump ignores HTTP proxy environment variables


jonathanfoster:
It appears the Splunk pump client isn't using HTTP proxy environment variables. I'm running Tyk Pump in a restricted environment that requires all outbound connections use the an HTTP proxy. Requests to the Splunk collector URL are timing out because they're attempting to connect directly instead of using the proxy. I confirmed HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables are all set.

I believe this could be caused by [setting the HTTP default client transport here](https://github.com/TykTechnologies/tyk-pump/blob/aa0961a3e2c40a5aa763a15a641b1651b72c770c/pumps/splunk.go#L305) and not setting proxy to `http.ProxyFromEnvironment`.

caroltyk:
@jonathanfoster Just an update we're working on it and will fix this in our next release.

vverbani:
@jonathanfoster came across this comment from the issue #4102... checked the internal ticket associated with this and it appears this was pushed to pump v1.10.0

PR #818, opened 2024-05-02
TT-12072 Configure Splunk client to use proxy from environment

jonathanfoster:
This PR resolves issue #816 by configuring the Splunk client to use proxy settings from the environment.

## Description

I set the proxy to `http.ProxyFromEnvironment` when initializing the Splunk client transport.

## Related Issue

- #816 

## Motivation and Context

Support for HTTP_PROXY environment variables is required when running Tyk Pump in restricted environments that require all HTTP traffic to go through a proxy.

## How This Has Been Tested

I included a unit test that verifies the proxy is not nil after initializing a Splunk client. I attempted to fully test the functionality by setting the HTTP_PROXY environment variable, but this caused other Splunk tests to fail. Since this fix leverages existing functionality from `http.ProxyFromEnvironment`, I don't think it's necessary to have end-to-end test coverage. 

Technically one could set Proxy to something other than ProxyFromEnvironment and the test would pass so this is a gap. It would be ideal to verify the function pointers are the same, but I couldn't get it to work. I'm open to feedback on how to better handle this test.

Not applicable.

## Types of changes

## Checklist

- [X] I have updated the documentation accordingly.
- [X] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [X] I have added tests to cover my changes.
- [X] Check your code additions will not fail linting checks:
  - [X] `go fmt -s`
  - [X] `go vet`

### DIFF
--- a/pumps/splunk.go
+++ b/pumps/splunk.go
@@ -302,7 +302,7 @@ func NewSplunkClient(token string, collectorURL string, skipVerify bool, certFil
 		}
 		tlsConfig = &tls.Config{Certificates: []tls.Certificate{cert}, ServerName: serverName}
 	}
-	http.DefaultClient.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+	http.DefaultClient.Transport = &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: tlsConfig}
 	// Append the default collector API path:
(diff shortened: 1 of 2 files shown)


jonathanfoster:
Issue fixed in https://github.com/TykTechnologies/tyk-pump/pull/820

PR #819, opened 2024-05-03
[SYSE-359]: Use distroless static as base image for tyk-pump

asutosh:
## Description

To avoid continued CVE hunting on docker images, try experimenting with distroless/static as base image as pump is built with `CGO_ENABLED=0`


alephnull:
Template based pr elsewhere.

PR #833, opened 2024-06-06
Removing FIPS from NFPMS

mativm02:
Removing FIPS from NFPMS to avoid this issue when creating a tag:
https://github.com/TykTechnologies/tyk-pump/actions/runs/9406734174/job/25910859694


alephnull:
Previously, the passphrase needed to be passed in a [separate env variable](https://goreleaser.com/customization/nfpm/#signing-key-passphrases) for each nfpm build. So we used `NFPM_STD_PASSPHRASE` in https://github.com/TykTechnologies/tyk-pump/blob/TT-12103-4/.github/workflows/release.yml#L93. Changing this to `NFPM_PASSPHRASE` should make it work for the fips build too.

Makefile
```diff
@@ -0,0 +1,13 @@
+build-fips:
+	GOEXPERIMENT=boringcrypto go build -tags=boringcrypto
+
+clean:
+	rm -f tyk-pump
+
+run-fips: build-fips
+	./$(BINARY_NAME)
```
sredxny, line 8: which value does it takes `BINARY_NAME` and where does this comes from? I guess we need to initialize it?

sredxny (commented):
overall it looks good! I just left a question/change let me know what do you think @mativm02

Makefile
mativm02: @sredxny it was an old variable, I forgot to remove it! thanks :)

sredxny (approved):
lgtm

PR #861, opened 2024-11-21
TT-13421 create indexes on sharded sql pumps

sredxny:
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/sql.go
+++ b/pumps/sql.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/sirupsen/logrus"
 	"sync"
 
 	"github.com/TykTechnologies/tyk-pump/analytics"
@@ -393,11 +394,18 @@ func (c *SQLPump) createIndex(indexBaseName, tableName, column string, wg *sync.
 	sql := fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS %s ON %s (%s)", option, indexName, tableName, column)
 	err := c.db.Exec(sql).Error
 	if err != nil {
-		c.log.Errorf("error creating index %s for table %s : %s", indexName, tableName, err.Error())
+		c.log.WithFields(logrus.Fields{
+			"index": indexName,
+			"table": tableName,
+		}).WithError(err).Error("Error creating index")
 		return err
 	}
 
 	c.log.Infof("Index %s created for table %s", indexName, tableName)
+	c.log.WithFields(logrus.Fields{
+		"index": indexName,
+		"table": tableName,
+	}).Info("Index created")
 	return nil
 }
 
--- a/pumps/sql.go
+++ b/pumps/sql.go
@@ -231,8 +231,19 @@ func (c *SQLPump) WriteData(ctx context.Context, data []interface{}) error {
 
 			table := analytics.SQLTable + "_" + recDate
 			c.db = c.db.Table(table)
+			fakerecord := &analytics.AnalyticsRecord{}
+			fakerecord.CollectionName = table
 			if !c.db.Migrator().HasTable(table) {
-				c.db.AutoMigrate(&analytics.AnalyticsRecord{})
+				err := c.db.AutoMigrate(&fakerecord)
+				if err != nil {
+					fmt.Println("Error creating table: ", err)
+				} else {
+					fmt.Println("Ensuring indexes for: ", table)
+					err = c.ensureIndex(table, true)
+					if err != nil {
+						fmt.Println("Error ensuring index: ", err)
+					}
+				}
 			}
 		} else {
 			i = dataLen // write all records at once for non-sharded case, stop for loop after 1 iteration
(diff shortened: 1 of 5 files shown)


pumps/sql.go
```diff
@@ -354,3 +370,82 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+// ensureIndex check that all indexes for the analytics SQL table are in place
+func (c *SQLPump) ensureIndex(tableName string, background bool) error {
+	if !c.db.Migrator().HasTable(tableName) {
+		return errors.New("cannot create indexes as table doesn't exist: " + tableName)
+	}
+
+	// waitgroup to facilitate testing and track when all indexes are created
+	var wg sync.WaitGroup
+	wg.Add(len(indexes))
+	createIndexFn := func(indexBaseName, column string) error {
```
jeffy-mathew, line 387: nit: for better readability, shall we move this func definition out of `ensureIndex`?

pumps/sql.go
```diff
@@ -354,3 +370,82 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+		}
+
+		sql := fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS %s ON %s (%s)", option, indexName, tableName, column)
+		err := c.db.Exec(sql).Error
+		if err != nil {
+			c.log.Errorf("error creating index %s for table %s : %s", indexName, tableName, err.Error())
+			return err
+		}
+
+		c.log.Infof("Index %s created for table %s", indexName, tableName)
```
jeffy-mathew, line 407: nit:
```suggestion
		c.log.WithFields(logrus.Fields{
			"index": indexName,
			"table": tableName,
		}).Info("index created")
```

pumps/sql.go
```diff
@@ -354,3 +370,82 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+
+		columnExist := c.db.Migrator().HasColumn(&analytics.AnalyticsRecord{}, column)
+		if !columnExist {
+			return errors.New("cannot create index for non existent column " + column)
+		}
+
+		sql := fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS %s ON %s (%s)", option, indexName, tableName, column)
+		err := c.db.Exec(sql).Error
+		if err != nil {
+			c.log.Errorf("error creating index %s for table %s : %s", indexName, tableName, err.Error())
```
jeffy-mathew, line 403: nit:
```suggestion
			c.log.WithFields(logrus.Fields{
				"index": indexName,
				"table": tableName,
			}).WithError(err).Error("error creating index")
```

pumps/sql.go
```diff
@@ -354,3 +370,82 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+		}
+
+		c.log.Infof("Index %s created for table %s", indexName, tableName)
+		return nil
+	}
+
+	for _, idx := range indexes {
+		indexName := tableName + idx.baseName
+
+		if !c.db.Migrator().HasIndex(tableName, indexName) {
```
jeffy-mathew, line 414: let's use early exits for better readability and to avoid nesting
```suggestion
		if c.db.Migrator().HasIndex(tableName, indexName) {
		  c.log.WithFields(logrus.Fields{
				"index": indexName,
				"table": tableName,
			}).Info("index already exists")
		continue
		}
		if background {...
```

pumps/sql.go
```diff
@@ -354,3 +370,82 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+}
+
+// ensureIndex check that all indexes for the analytics SQL table are in place
+func (c *SQLPump) ensureIndex(tableName string, background bool) error {
+	if !c.db.Migrator().HasTable(tableName) {
+		return errors.New("cannot create indexes as table doesn't exist: " + tableName)
+	}
+
+	// waitgroup to facilitate testing and track when all indexes are created
+	var wg sync.WaitGroup
```
jeffy-mathew, line 385: do we need waitgroup  when create background is false?

jeffy-mathew:
@sredxny since we're using `gorm`, shall we explore the possibility of using [NamingStrategy](https://gorm.io/docs/gorm_config.html#NamingStrategy) in gorm config?

pumps/sql.go
```diff
@@ -354,3 +372,99 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+		if c.db.Migrator().HasIndex(tableName, indexName) {
+			c.log.WithFields(logrus.Fields{
+				"index": indexName,
+				"table": tableName,
+			}).Info("Index already exists")
+			continue
+		}
+
+		if background {
+			go func(baseName, cols string) {
```
jeffy-mathew, line 438: nit: shall we make `createIndex` waitgroup free, and limit the goroutine + wg usage in this anon block?
```suggestion
			go func(baseName, cols string) {
				defer wg.Done()
				if err := c.createIndex(baseName, tableName, cols); err != nil {
					c.log.Error(err)
				}
			}(idx.baseName, idx.column)
```

pumps/sql.go
```diff
@@ -354,3 +372,96 @@ func (c *SQLPump) WriteUptimeData(data []interface{}) {
...
+	if c.dbType == "postgres" {
+		option = "CONCURRENTLY"
+	}
+
+	columnExist := c.db.Migrator().HasColumn(&analytics.AnalyticsRecord{}, column)
+	if !columnExist {
+		return errors.New("cannot create index for non existent column " + column)
+	}
+
+	sql := fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS %s ON %s (%s)", option, indexName, tableName, column)
```
mativm02, line 392: A better naming for this variable could be `query`, what do you think?

pumps/sql_aggregate.go
```diff
@@ -198,7 +198,7 @@ func (c *SQLAggregatePump) ensureIndex(tableName string, background bool) error
 		c.log.Info("Creating index for table ", tableName, "...")
 		return createIndexFn(c)
 	}
-	c.log.Info(newAggregatedIndexName, " already exists.")
+	c.log.Info(indexName, " already exists.")
```
mativm02, line 201: will this be printed every time we run pump and the indexes were already created? If so, this could create a lot of noise

pumps/sql.go
jeffy-mathew: In addition to this, let's bring in practise to be careful when forming SQL queries. I know the values in this query are totally under our control (consts), but still I feel like it's better to use an sql builder. https://gorm.io/docs/sql_builder.html - this might help

pumps/sql_aggregate.go
sredxny: thats true, we dont need to print that. And we are already printing errors inside the `createIndexFn` so am removing it

pumps/sql.go
sredxny: Renamed.... @jeffy-mathew in such case we would need to add some logic like:

```
query := "CREATE INDEX IF NOT EXISTS ? ON ? (?)"
if option != "" {
    query = fmt.Sprintf("CREATE INDEX %s IF NOT EXISTS ? ON ? (?)", option) // Add option dynamically
}
```

As the original query contains a param "option" that can be "concunrrently" or empty, and db.Raw method doesnt allows it and will mark it as bad query

PR #865, opened 2025-01-24
[TT-13937] Adding key obfuscation to the Prometheus pump

vmihaylov3:
Parameters for api_key obfuscation are added to the Prometheus pump.

## Description
- Adding 2 parameters in the Prometheus custom metrics in which the user can select if they want to obfuscate the API key and if yes, how many characters they want to show in it.
- Adding a validation in the `GetLabelsValues` function of the promethus pump that will validate if the variables are set and it will modify the api_key and the key parameters.

## Related Issue
N/A

## Motivation and Context
- In my case the /metrics endpoint that is exposed by the container can be accessed by other user. We need to obfuscate the key so that we can reduce the chance of a key being leaked.

## How This Has Been Tested
- Tested on a remote env using applications that are making constant requests to the gateways. In this env, we are using MongoDB and Redis.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If pulling from your own
      fork, don't request your `master`!
- [x] Make sure you are making a pull request against the **`master` branch** (left side). Also, you should start
      *your branch* off *our latest `master`*.
- [x] My change requires a change to the documentation.
  - [ ] If you've changed APIs, describe what needs to be updated in the documentation.
- [x] I have updated the documentation accordingly.
- [ ] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] Check your code additions will not fail linting checks:
  - [ ] `go fmt -s`
  - [ ] `go vet`

PR #866, opened 2025-01-28
[TT-13937] Adding key obfuscation to the Prometheus pump

mativm02:
Duplicate of #865 
Parameters for api_key obfuscation are added to the Prometheus pump.

## Description
- Adding 2 parameters in the Prometheus custom metrics in which the user can select if they want to obfuscate the API key and if yes, how many characters they want to show in it.
- Adding a validation in the `GetLabelsValues` function of the promethus pump that will validate if the variables are set and it will modify the api_key and the key parameters.

## Related Issue
https://tyktech.atlassian.net/jira/software/c/projects/TT/boards/42?selectedIssue=TT-13937

## Motivation and Context
- In my case the /metrics endpoint that is exposed by the container can be accessed by other user. We need to obfuscate the key so that we can reduce the chance of a key being leaked.

## How This Has Been Tested
- Tested on a remote env using applications that are making constant requests to the gateways. In this env, we are using MongoDB and Redis.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/prometheus.go
+++ b/pumps/prometheus.go
@@ -67,6 +67,12 @@ type PrometheusMetric struct {
 	MetricType string `json:"metric_type" mapstructure:"metric_type"`
 	// Defines the buckets into which observations are counted. The type is float64 array and by default, [1, 2, 5, 7, 10, 15, 20, 25, 30, 40, 50, 60, 70, 80, 90, 100, 200, 300, 400, 500, 1000, 2000, 5000, 10000, 30000, 60000]
 	Buckets []float64 `json:"buckets" mapstructure:"buckets"`
+	// Controls whether the pump client should hide the API key. In case you still need substring
+	// of the value, check the next option. Default value is `false`.
+	ObfuscateAPIKeys bool `json:"obfuscate_api_keys" mapstructure:"obfuscate_api_keys"`
+	// Define the number of the characters from the end of the API key. The `obfuscate_api_keys`
+	// should be set to `true`. Default value is `0`.
+	ObfuscateAPIKeysLength int `json:"obfuscate_api_keys_length" mapstructure:"obfuscate_api_keys_length"`
 	// Defines the partitions in the metrics. For example: ['response_code','api_name'].
 	// The available labels are: `["host","method",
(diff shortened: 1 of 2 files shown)


pumps/prometheus.go
```diff
@@ -67,6 +67,12 @@ type PrometheusMetric struct {
 	MetricType string `json:"metric_type" mapstructure:"metric_type"`
 	// Defines the buckets into which observations are counted. The type is float64 array and by default, [1, 2, 5, 7, 10, 15, 20, 25, 30, 40, 50, 60, 70, 80, 90, 100, 200, 300, 400, 500, 1000, 2000, 5000, 10000, 30000, 60000]
 	Buckets []float64 `json:"buckets" mapstructure:"buckets"`
+	// Controls whether the pump client should hide the API key. In case you still need substring
+	// of the value, check the next option. Default value is `false`.
+	ObfuscateAPIKeys bool `json:"obfuscate_api_keys" mapstructure:"obfuscate_api_keys"`
+	// Define the number of the characters from the end of the API key. The `obfuscate_api_keys`
+	// should be set to `true`. Default value is `4`.
+	ObfuscateAPIKeysLength int `json:"obfuscate_api_keys_length" mapstructure:"obfuscate_api_keys_length"`
```
sredxny, line 75: I think we can remove this config, if we can calculate how many digits to display/hide would be the best, or, we can just set arbitrary number of digits in the same way as we do in the gateway, check: https://github.com/TykTechnologies/tyk/blob/master/gateway/log_helpers.go#L14

what do you think? is there any special case for which you think we should keep it as a config?

pumps/prometheus.go
```diff
@@ -400,6 +408,21 @@ func (pm *PrometheusMetric) GetLabelsValues(decoded analytics.AnalyticsRecord) [
 	return values
 }
 
+func obfuscateAPIKey(apiKey string, obfuscate bool, length int) string {
```
sredxny, line 411: also, if we only use this function in prometheus can we just change the signature to something like:

`func(pm *PrometheusMetric) obsfusctaeKey(key string){}` ?

sredxny (changes requested):

sredxny (approved):
✅

PR #870, opened 2025-02-20
[TT-13341] Remove SQLite support from Tyk Pump

mativm02:
## Description

This PR removes SQLite as a supported SQL pump type across the project:
- Updated documentation to remove SQLite references
- Removed SQLite-related dependencies from go.mod and go.sum
- Updated test cases to use Postgres instead of SQLite
- Updated configuration documentation to clarify supported SQL pump types
- Removed SQLite-specific code in SQL pump implementations

## Related Issue

https://tyktech.atlassian.net/browse/TT-13341
## Motivation and Context

https://tyktech.atlassian.net/browse/TT-13341
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/pumps/graph_sql_aggregate_test.go
+++ b/pumps/graph_sql_aggregate_test.go
@@ -74,6 +74,9 @@ const (
 )
 
 func TestSqlGraphAggregatePump_Init(t *testing.T) {
+	if os.Getenv("TYK_TEST_POSTGRES") == "" {
+		t.Skip("Skipping test because TYK_TEST_POSTGRES environment variable is not set")
+	}
 	tableName := analytics.AggregateGraphSQLTable
 	r := require.New(t)
 	pump := &GraphSQLAggregatePump{}
@@ -81,7 +84,7 @@ func TestSqlGraphAggregatePump_Init(t *testing.T) {
 		conf := SQLAggregatePumpConf{
 			SQLConf: SQLConf{
 				Type:             "postgres",
-				ConnectionString: "",
+				ConnectionString: getTestPostgresConnectionString(),
 			},
 		}
 		assert.NoError(t, pump.Init(conf))
@@ -115,8 +118,9 @@ func TestSqlGraphAggregatePump_Init(t *testing.T) {
 
 	t.Run("decode from map", func(t *testing.T) {
 		conf := map[string]interface{}{
-			"type":           "postgres",
-			"table_sharding": true,
+			"type":              "postgres",
+			"table_sharding":    true,
+			"connection_string": getTestPostgresConnectionString(),
 		}
 		r.NoError(pump.Init(conf))
 		assert.Equal(t, "postgres", pump.SQLConf.Type)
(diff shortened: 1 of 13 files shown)


pumps/graph_sql_aggregate_test.go
```diff
@@ -74,14 +74,17 @@ const (
 )
 
 func TestSqlGraphAggregatePump_Init(t *testing.T) {
+	if os.Getenv("TYK_TEST_POSTGRES") == "" {
+		t.Skip("Skipping test because TYK_TEST_POSTGRES environment variable is not set")
+	}
```
sredxny, line 79: what about if we move this to a helper function? in sink we have:

```
func ValidTestType(t *testing.T, storageType config.DBType) {
	t.Helper()
	config.LoadConfig("")

	if config.GetConfig().AnalyticsConfig.Type == storageType {
		return
	} else {
		t.Skip()
	}
}
```
wondering if we should something similar here

pumps/sql.go
```diff
@@ -58,7 +57,8 @@ type SQLConf struct {
 	// The prefix for the environment variables that will be used to override the configuration.
 	// Defaults to `TYK_PMP_PUMPS_SQL_META`
 	EnvPrefix string `mapstructure:"meta_env_prefix"`
-	// The supported and tested types are `sqlite` and `postgres`.
+	// The only supported and tested types are `postgres` and `mysql`.
```
sredxny, line 60: maybe we should start the deprecation process for mysql as well

sredxny: as we do not really test with mysql

pumps/sql_test.go
```diff
@@ -446,6 +469,7 @@ func TestEnsureIndexSQL(t *testing.T) {
 			assert.NotNil(t, pmp)
 
 			actualErr := pmp.ensureIndex(tc.givenTableName, tc.givenRunInBackground)
+			fmt.Println("actualErr", actualErr)
```
sredxny, line 472: remove

pumps/sql_test.go
```diff
@@ -363,10 +383,11 @@ func TestSQLWriteUptimeDataAggregations(t *testing.T) {
 }
 
 func TestDecodeRequestAndDecodeResponseSQL(t *testing.T) {
+	skipTestIfNoPostgres(t)
 	newPump := &SQLPump{}
 	cfg := make(map[string]interface{})
-	cfg["type"] = "sqlite"
-	cfg["connection_string"] = ""
+	cfg["type"] = "postgres"
+	cfg["connection_string"] = getTestPostgresConnectionString()
```
sredxny, line 390: I see this repeated many times, what about creating a kind of constructor for this? something like:

```
NewSQLConfig(shard bool)map[string]interface{}{
        cfg := make(map[string]interface{})
	cfg["type"] = "postgres"
	cfg["connection_string"] = getTestPostgresConnectionString()
	cfg["table_sharding"] = shard
	return cfg
}
```

pumps/graph_sql_aggregate_test.go
sredxny: I see down below you already have: `skipTestIfNoPostgres` lets use that

sredxny (changes requested):

pumps/graph_sql_aggregate_test.go
mativm02: I forgot to add skipTestIfNoPostgres there

sredxny (approved):

PR #884, opened 2025-08-14
Fix syslog pump log fragmentation issue

buger:
## Summary
- Fix log fragmentation in syslog pump when raw_request/raw_response contain newlines
- Serialize analytics data to JSON before writing to syslog writer
- Ensures entire record is treated as single atomic message

## Problem
The syslog pump was causing log fragmentation when `raw_request` and `raw_response` fields contained newline characters. The `fmt.Fprintf` call on line 171 was not properly escaping these characters, causing the syslog daemon to interpret each newline as a separate log entry, breaking single analytics records into multiple fragmented log entries.

## Solution
- Added `encoding/json` import
- Marshal the message map to JSON before writing to syslog
- Added proper error handling for JSON marshaling failures
- This ensures the raw request/response data is properly escaped within the JSON structure

## Test plan
- [x] Verify the code compiles successfully
- [ ] Test with syslog pump configuration that includes raw request/response data containing newlines
- [ ] Verify that analytics records appear as single log entries in syslog output

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pumps/syslog.go
+++ b/pumps/syslog.go
@@ -2,6 +2,7 @@ package pumps
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/syslog"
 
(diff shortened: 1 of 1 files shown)


buger:
@sredxny @mativm02 false start. See varmir message. We should be very careful during the PR review. Plus it had no any tests to it, can't be approved 😛

PR #962, opened 2026-04-14
TT-16964 - remove sbom job from release workflow

bsten-tyk:
## Description

This PR fixes three CI/CD issues introduced by commit 339740c ("security: pin actions to SHA, harden CI permissions, pin Docker images" #956) that have broken both the `ci-test.yml` and `release.yml` workflows on master since April 1, 2026.

**Changes:**

1. **Remove `sbom` job from `release.yml`** -- The `sbom` job calls a reusable workflow that requests `id-token: write`, but the workflow-level `permissions` block only grants `contents: read`. This caused all release workflow runs on master to fail with `startup_failure`.

2. **Fix `goimports` install in `ci-test.yml`** -- The security hardening changed `go install` to `go run` for goimports. `go run` executes the tool once but does not install the binary to PATH, causing `ci-test.sh` to fail with "goimports: command not found". Restored to `go install` with a pinned version (`@v0.33.0`).

3. **Fix multi-platform manifest digests in `Dockerfile.distroless`** -- Base images were pinned to single-platform (amd64-only) SHA256 digests instead of multi-platform manifest list digests. This broke arm64 and s390x Docker builds because `dpkg` rejected architecture mismatches. Replaced with manifest list digests that cover all required platforms.

## Related Issue

TT-16964

## Motivation and Context

Both CI workflows (`ci-test.yml` and `release.yml`) have been broken on master since April 1, 2026 due to unintended side effects from the security hardening commit. No successful release workflow run has occurred since March 26. This PR restores CI functionality while preserving the security intent (pinned versions and digests).

## How This Has Been Tested

- The `sbom` permission error is a workflow validation failure at parse time -- removing the job eliminates it.
- The `goimports` fix was verified by analyzing the CI failure logs showing "command not found" at `ci-test.sh:33`.
- The Dockerfile digest fix was verified using `docker buildx imagetools inspect` to confirm manifest list digests cover amd64, arm64, and s390x platforms.

N/A

## Types of changes

## Checklist

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side).
- [x] Make sure you are making a pull request against the **`master` branch** (left side).
  - [x] `go fmt -s`
  - [x] `go vet`


ci/Dockerfile.distroless
```diff
@@ -10,7 +10,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 COPY ${BUILD_PACKAGE_NAME}_*${TARGETARCH}.deb /
 RUN dpkg -i /${BUILD_PACKAGE_NAME}_*${TARGETARCH}.deb && rm /*.deb
 
-FROM gcr.io/distroless/static-debian12:nonroot@sha256:5074667eecabac8ac5c5d395100a153a7b4e8426181cca36181cd019530f00c8
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:a9329520abc449e3b14d5bc3a6ffae065bdde0f02667fa10880c49b35c109fd1
```
asutosh: I don't think we can add sha here, we depend on this step to get the latest debian security fix during the build time for all our releases.
If we pin to a sha here, it would break this workflow, and we'll no longer get security fixes, unless we update the image digests for every release, and all our CVE timelines/SLAs for debian fixes would break.
I think it could work with an associated workflow with dependabot or digestabot to update digests, but these PRs should still be approved, merged and picked to the right branches to get the latest updates.
Until we have such similar steps set up in practice, I don't think we can use sha pinning here, and will have to accept this risk, and trust google/debian that the upstream images pulled using tags are safe.

bsten-tyk: okay, updated tags in c2444a01145c191fa7d9d7eedd6735f7fc5a626e based on your suggestion.

ci/Dockerfile.distroless
```diff
@@ -10,7 +10,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 COPY ${BUILD_PACKAGE_NAME}_*${TARGETARCH}.deb /
 RUN dpkg -i /${BUILD_PACKAGE_NAME}_*${TARGETARCH}.deb && rm /*.deb
 
-FROM gcr.io/distroless/static-debian12:nonroot@sha256:5074667eecabac8ac5c5d395100a153a7b4e8426181cca36181cd019530f00c8
+FROM gcr.io/distroless/static-debian12:nonroot
```
asutosh, line 13: We can move to debian 13 here, as with nonroot static images, there shouldn't be no issue with bumping debian version. (No GLIBC)

asutosh (approved):

PR #989, opened 2026-04-27
[TT-17004] Fix MCP Mongo aggregate cross-API merge

andrei-tyk:
## Summary

The `MCPMongoAggregatePump` upsert query only matched on `(orgid, timestamp)`, so concurrent writes from two MCP proxies in the same time bucket merged into a single document. Dashboard queries scoped to a single api (`apiid.<X>: \$exists`) still matched the merged doc; unwinding `lists.names` returned counters from BOTH proxies; queries reported the sum (e.g. 5+2=7) instead of one proxy's traffic.

## Fix

- Add `OwnerAPIID` (bson `owner_apiid`) to `MCPRecordAggregate`.
- Populate it in `initMCPAggregateForRecord` from `record.APIID`.
- Include it in the upsert query in `pumps/mcp_mongo_aggregate.go` so each `(org, timestamp, api)` gets its own document.

Each MCP doc now holds a single api's `lists.names` entries and the existing dashboard queries are correct.

## Test plan

- Reproduced by tyk-analytics test `TestMCPAnalyticsProxyFilter::test_api_id_filter_scopes_results_to_selected_proxy` (added in tyk-analytics PR #5514).
- Existing pump unit tests in `analytics/aggregate_mcp_test.go` and `pumps/mcp_mongo_aggregate_test.go` continue to pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/analytics/aggregate_mcp.go
+++ b/analytics/aggregate_mcp.go
@@ -14,8 +14,17 @@ const (
 // MCPRecordAggregate holds aggregated MCP analytics grouped by API.
 // It embeds AnalyticsRecordAggregate for all standard dimensions and adds
 // MCP-specific dimension maps for method, primitive type, and primitive name.
+//
+// OwnerAPIID identifies which API this aggregate belongs to. It partitions
+// MongoDB documents per (org, timestamp, api) so per-api Names/Methods/
+// Primitives counters from one proxy don't merge with another's via the
+// upsert in MCPMongoAggregatePump. The embedded AnalyticsRecordAggregate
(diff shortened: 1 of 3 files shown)


lghiur (approved):

andrei-tyk:
api-tests were validated to be green on tyk-analytics after a run with the tyk-pump pointing at this commit.
https://github.com/TykTechnologies/tyk-analytics/actions/runs/25023463623/job/73292902189?pr=5514

PR #1058, opened 2026-07-30
TT-9424: Sort aggregate batches by ID to prevent PostgreSQL deadlocks

nerdydread:
## Description

Multi-replica Pump deployments writing analytics aggregates to PostgreSQL deadlock on the
`tyk_aggregated*` tables with `ERROR: deadlock detected (SQLSTATE 40P01)`.

The cause is nondeterministic row ordering within a multi-row `INSERT … ON CONFLICT (id) DO UPDATE`.
`Dimensions()` builds the record slice by ranging over Go maps, Go randomises map iteration per
process, and PostgreSQL acquires row locks in statement order. Two replicas upserting overlapping
IDs in opposite orders form a circular wait, and PostgreSQL aborts one of them.

The fix is deterministic lock ordering — the resource-ordering pattern. Each batch is sorted by row
ID before the upsert, so every replica requests locks in the same sequence and contending
transactions queue instead of cycling:

```go
sort.Slice(recs, func(i, j int) bool { return recs[i].ID < recs[j].ID })
```

Applied at all four bulk-upsert call sites. `rg -n 'clause\.OnConflict|OnConflictAssignments' pumps/`
returns exactly these four non-test occurrences:

| File | Function | Sort at |
| --- | --- | --- |
| `pumps/sql_aggregate.go` | `DoAggregatedWriting` | 302 |
| `pumps/graph_sql_aggregate.go` | `DoAggregatedWriting` | 179 |
| `pumps/mcp_sql_aggregate.go` | `DoAggregatedWriting` | 256 |
| `pumps/sql.go` | `WriteUptimeData` | 408 |

All four upsert on an `id` arbiter whose value is a hex-encoded string, so lexicographic comparison
matches the arbiter exactly.

In `pumps/sql.go` the placement is load-bearing. `recs` is rebuilt per organisation and per
day-shard, so the sort sits in the innermost scope, immediately before the batching loop. Hoisting
it above either enclosing loop would leave later orgs and later shards unsorted.

**Three things to note:**

1. **This changes physical row insertion order.** No API or schema change, but anything relying on
   the order of an unordered `SELECT` will shift. Exactly one existing test did:
   `TestSQLWriteUptimeDataAggregations` called `Find()` with no `ORDER BY` and asserted on
   `dbRecords[0]`, relying on heap order. After sorting, index 0 became the `errors` dimension row
   instead of `url`. The assertion was unsound rather than the fix being wrong, so it now selects
   its row by dimension. Nothing else in the suite depended on the old ordering.

2. **Sorting also makes batch boundaries deterministic.** Because `recs` is chunked by `BatchSize`,
   replica A's batch *n* and replica B's batch *n* now cover an identical ID set rather than
   partially overlapping ones, removing cross-batch contention as well as intra-statement lock
   disorder.

3. **One unrelated test change is included, to unblock CI.** `pumps/mcp_sql_aggregate_test.go`
   gains `OmitIndexCreation: true` in one test config. This fixes a **pre-existing test failure that is
   not caused by this PR** — master fails it with a byte-identical signature (run
   [29005522096](https://github.com/TykTechnologies/tyk-pump/actions/runs/29005522096), same
   subtest, same three missing records, same line 221). Detail in the testing section below.
   Happy to split it into its own PR if reviewers would prefer.

Alternatives considered and rejected as disproportionate: leader election or distributed locking
(bottlenecks all aggregate writes through one instance) and workload partitioning (requires
consistent hashing, topology awareness, and rebalancing). The sort is `O(N log N)` over a few
hundred in-memory elements and remains correct under either approach if one is later adopted.

## Related Issue

TT-9424 — https://tyktech.atlassian.net/browse/TT-9424

Reported against Pump 1.13.2 and still live in production. Affects JAMF and Kingfisher.

## Motivation and Context

Each deadlock aborts the transaction, and `DoAggregatedWriting` returns without retry — so the
purge cycle's aggregates are lost. This is **silent analytics data loss**, not just log noise, which
is why it ships as a bug fix rather than a hardening task.

Under the reproduced workload the loss rate is severe: 24 of 45 write cycles (53%) failed on
unmodified master.

The only workaround available to customers today is dropping to a single Pump replica, which gives
up horizontal scaling and HA. All four affected paths are fixed together rather than just the
reported one — the edit is mechanically identical in each, they share the same defect for the same
reason, and a partial fix would leave three paths still dropping analytics data while producing a
confusing support picture where deadlocks vanish from one table and persist on others.

## How This Has Been Tested

New regression suite: `pumps/sql_deadlock_ordering_test.go`. It stands up three independently
initialised pump instances, each with its own database connection, mirroring separate pods, which
then concurrently write an overlapping batch.

```
TYK_TEST_POSTGRES="host=localhost port=5433 user=postgres password=test \
  dbname=tyk_analytics sslmode=disable" \
  go test ./pumps -run TestTT9424 -v -count=1
```

Environment: PostgreSQL 16 (`postgres:16-alpine`, Docker), macOS, 3 simulated replicas × 15 write
cycles.

**Every test was verified as a real guard** — each was confirmed to *fail* with the `sort.Slice`
calls reverted, then pass with them applied. A test that passes both ways guards nothing.

| Test | Without fix | With fix |
| --- | --- | --- |
| SQL aggregate pump, non-sharded | 24 deadlocks / 45 cycles | 0 |
| SQL aggregate pump, sharded | 26 deadlocks / 45 cycles | 0 |
| Graph aggregate pump | 24 deadlocks / 45 cycles | 0 |
| MCP aggregate pump | 26 deadlocks / 45 cycles | 0 |
| Uptime pump | batches lost, hit counts short | 0, counts exact |
| Sort determinism | n/a (property test) | pass |

Sharded coverage is included because the customer reports reference the dated
`tyk_aggregated_YYYYMMDD` tables specifically.

The uptime test writes several organisations per cycle, which is what covers the placement
constraint above: `recs` is rebuilt per org, so a sort hoisted out of the per-org loop leaves every
org after the first unsorted and deadlocking.

The determinism test asserts the property the fix depends on directly — identical input produces
identical row order and identical batch boundaries across runs — rather than inferring it from the
absence of deadlocks.

**A note for anyone modifying the fixtures.** Reproduction is very sensitive to workload shape, and
not in the intuitive direction: adding more data makes the bug reproduce *less*, because splitting
`recs` across batches turns each batch into its own short transaction. Measured against an unfixed
tree, three replicas, graph pump:

| Workload | Cycles deadlocked (no fix) |
| --- | --- |
| 749 rows from 4 dimension maps | 0 / 45 |
| 1310 rows from 7 dimension maps | 24 / 45 |
| 2001 rows from 4 dimension maps | 2 / 120 |
| 3501 rows from 7 dimension maps | 3 / 120 |

The fixtures sit on the 1310-row shape. This is recorded in a header comment in the test file with a
warning to re-verify against an unfixed tree before changing any cardinality or batch size,
otherwise the suite can quietly stop guarding anything.

**Tests that run without PostgreSQL.** The deadlock tests above are necessarily Postgres-backed,
and the coverage CI job provisions only redis and mongo — so `skipTestIfNoPostgres` skipped them
there, leaving two of the four new `sort.Slice` lines uncovered and failing the 80% new-code
quality gate at 50%. Two SQLite-backed tests close that gap:

- `TestGraphSQLAggregatePump_DoAggregatedWriting_SQLite` (`pumps/graph_sql_aggregate_test.go`)
- `TestSQLWriteUptimeData_SQLite` (`pumps/sql_test.go`)

Both follow the existing `newMCPSQLAggregatePumpWithSQLite` pattern, reusing the
`setupTestDBWithJSONTags` helper and building the pump struct directly — `Init()` is not usable
here because `Dialect()` has not supported sqlite since v1.12.0. The other two sort sites were
already reachable from existing SQLite tests. New-code coverage measured without Postgres is now
**4/4 sort lines**, up from 2/4.

These deliberately do not assert row ordering. Proving insertion order through an unordered
`Find()` is the same unsound pattern fixed in `TestSQLWriteUptimeDataAggregations` above; the
ordering property is asserted directly, and without a database, by `TestTT9424SortDeterminism`.

**The pre-existing MCP test failure.** `TestMCPSQLAggregatePump_WriteData` fails intermittently on
master — roughly one CI run in three — and blocked this PR twice. It is not caused by this change:
master run 29005522096 fails it with an identical signature.

Mechanism, from the CI logs: `Init` spawns a background `CREATE INDEX CONCURRENTLY` goroutine that
outlives the subtest that started it. Its "index created successfully" line appears *inside the
next subtest*, and the subtest that fails is always the one immediately following a subtest that
took ~1s instead of ~0.02s — the signature of a blocked index build. When the goroutine collides
with the subtest's `DROP TABLE` cleanup the drop loses, its error is discarded, and the next
subtest's writes accumulate onto stale rows, so expected counts no longer match.

Setting `OmitIndexCreation: true` removes that goroutine for this test. Index creation is still
covered by the `Init` subtests and `TestEnsureIndexSQLAggregate`.

**This fix could not be verified locally** — the failure was not able to be reproduced locally
in either direction (0 failures in 12 runs both with and without the change), because a local
index build takes 0.02s and never spans into the next subtest like on the test runner. The change is justified by the
mechanism above rather than by local reproduction, and 12/12 green confirms only that nothing
regressed. CI is the real verdict, which now also passes.

**Wider regression run.** 214 non-Mongo tests in `./pumps` pass, as do all other packages
(`analytics`, `serializer`, `storage`, `logger`, root). `gofmt -s -l` and `go vet ./...` are clean
repo-wide.

The 55 Mongo-backed tests in `./pumps` could not run locally — the available Mongo is a replica set
advertising an internal hostname, so the host-side driver cannot reach it. None of them touch the
changed code paths, but **CI should be treated as the authority on those**.

N/A — no user-facing surface.

## Types of changes

## Checklist
  - [x] `go fmt -s`
  - [x] `go vet`

[TT-9424]: https://tyktech.atlassian.net/browse/TT-9424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-9424" title="TT-9424" target="_blank">TT-9424</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | [Innersource] SQL pump producing "error writing aggregated records deadlock detected (SQLSTATE 40P01) |

Generated at: 2026-07-30 23:50:37

</details>

### DIFF
--- a/pumps/graph_sql_aggregate.go
+++ b/pumps/graph_sql_aggregate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"sort"
 
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/mitchellh/mapstructure"
@@ -173,6 +174,9 @@ func (s *GraphSQLAggregatePump) DoAggregatedWriting(ctx context.Context, table, 
 		rec.Counter.ErrorMap = nil
 		recs = append(recs, rec)
 	}
+
+	// Deterministic lock ordering - see the note in SQLAggregatePump.DoAggregatedWriting (TT-9424).
+	sort.Slice(recs, func(i, j int) bool { return recs[i].ID < recs[j].ID })
 
 	for i := 0; i < len(recs); i += s.SQLConf.BatchSize {
 		ends := i + s.SQLConf.BatchSize
--- a/pumps/graph_sql_aggregate_test.go
+++ b/pumps/graph_sql_aggregate_test.go
@@ -669,3 +669,58 @@ func TestGraphSQLAggregatePump_WriteData_Sharded(t *testing.T) {
 		assert.NotEmpty(t, aggr, "table %s does not contain records", secondShardedTable)
 	})
 }
+
+// TestGraphSQLAggregatePump_DoAggregatedWriting_SQLite exercises the graph aggregate write
+// path against in-memory SQLite, so it runs everywhere - including the coverage CI job, which
+// provisions no PostgreSQL and therefore skips every Postgres-gated test in this package.
+//
+// Init() is deliberately not called: Dialect() has not supported sqlite since v1.12.0, so the
+// pump is built directly over the test database, mirroring newMCPSQLAggregatePumpWithSQLite.
+//
+// The deterministic batch ordering this path relies on (TT-9424) is asserted directly by
+// TestTT9424SortDeterminism and exercised under real contention by
+// TestTT9424GraphAggregateNoDeadlock; this test covers the write path itself.
+func TestGraphSQLAggregatePump_DoAggregatedWriting_SQLite(t *testing.T) {
+	db := setupTestDBWithJSONTags(t)
+	tableName := analytics.AggregateGraphSQLTable
+	require.NoError(t, db.Table(tableName).AutoMigrate(&analytics.GraphSQLAnalyticsRecordAggregate{}))
+
+	pump := &GraphSQLAggregatePump{
+		db: db,
+		SQLConf: &SQLAggregatePumpConf{
+			SQLConf: SQLConf{BatchSize: 100},
+		},
+	}
+	pump.log = log.WithField("prefix", SQLAggregatePumpPrefix)
+
+	ag := &analytics.GraphRecordAggregate{
+		Types:      map[string]*analytics.Counter{},
+		Fields:     map[string]*analytics.Counter{},
+		Operation:  map[string]*analytics.Counter{},
+		RootFields: map[string]*analytics.Counter{},
+	}
+	ag.TimeStamp = time.Date(2025, 6, 15, 10, 0, 0, 0, time.UTC)
+	ag.OrgID = "org1"
+	// Dimensions() always appends a "total" row built from this counter, and it must carry
+	// hits: the on-conflict assignment for request_time averages over
+	// (existing hits + incoming hits), which divides by zero when both sides are empty.
+	ag.Total = analytics.Counter{Hits: 4, Success: 4}
+	for i := 0; i < 4; i++ {
+		suffix := fmt.Sprintf("%02d", i)
+		ag.Types["Type"+suffix] = &analytics.Counter{Hits: 1, Success: 1}
+		ag.Fields["field"+suffix] = &analytics.Counter{Hits: 1, Success: 1}
+	}
+
+	require.NoError(t, pump.DoAggregatedWriting(context.Background(), tableName, "org1", "api1", ag))
+
+	var recs []analytics.GraphSQLAnalyticsRecordAggregate
+	require.NoError(t, pump.db.Table(tableName).Find(&recs).Error)
+
+	// 4 types + 4 fields + the implicit "total" row.
+	assert.Len(t, recs, 9)
+	for _, rec := range recs {
+		assert.Equal(t, "org1", rec.OrgID)
+		assert.Equal(t, "api1", rec.APIID)
+		assert.NotEmpty(t, rec.ID)
+	}
+}
--- a/pumps/mcp_sql_aggregate.go
+++ b/pumps/mcp_sql_aggregate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"sort"
 
 	"github.com/TykTechnologies/tyk-pump/analytics"
 	"github.com/mitchellh/mapstructure"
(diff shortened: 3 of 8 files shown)


nerdydread:
The logic issue reported on line 179 of `pumps/graphql_sql_aggregate.go` is not comparing an element to itself.  This is the line which already reflects the suggested fix: `sort.Slice(recs, func(i, j int) bool { return recs[i].ID < recs[j].ID })`

tbuchaillot (approved):
LGTM

nerdydread:
This was merged before QA sign-off. Reverting from `master` via #1066 so it can go through QA and be re-landed deliberately.

For anyone picking this up later: no release contains eead02ed (`git tag --contains` is empty), and CI was green on it — the revert is about restoring the QA gate, not fixing a break.

The head branch `TT-9424-sql-aggregate-deadlock` has been restored on origin at aa1f9136 so QA has a ref to build from. This PR cannot be reopened once #1066 merges, so the re-land will be a new PR that reverts the revert.

nerdydread:
Correction to my note above: the revert PR is now #1067, not #1066. #1066 was closed because its `revert-1058-*` branch name made the Jira linter resolve a nonexistent `REVERT-1058` ticket; #1067 carries the identical revert commit on a `TT-9424-` prefixed branch.

PR #1066, opened 2026-08-06
Revert "TT-9424: Sort aggregate batches by ID to prevent PostgreSQL deadlocks (#1058)"

nerdydread:
Reverts #1058 (merged as eead02ed).

#1058 was approved and merged before it had gone through QA. This reverts it from `master` so the change can be QAd and then re-landed deliberately.

### Context

- Not urgent and nothing is broken: `CI tests` and `Linter` both passed on eead02ed, and `git tag --contains eead02ed` is empty, so **no release contains this change** (latest tag is v1.16.0).
- eead02ed was a squash merge (single parent), so this is a plain `git revert` with no `-m` flag.
- Reverting rather than rewriting history: the `prevent-delete-force-push` ruleset protects `master`, and the commit is already public.

### Re-landing

The original branch `TT-9424-sql-aggregate-deadlock` has been restored on origin at aa1f9136 for QA to build from. Once QA signs off, the change will be re-landed via a new PR that reverts this revert (#1058 cannot be reopened once this merges).

### Verification

- Diff is the exact inverse of #1058: same 8 files, +9/-689.
- `go build ./...` passes.
- `go test ./pumps/ -run TestSQL -count=1` passes. The deadlock-ordering tests are removed along with the change, so their absence is expected rather than a regression.


nerdydread:
Superseded by #1067. The `revert-1058-*` branch name made the Jira linter resolve a nonexistent `REVERT-1058` ticket; #1067 carries the identical revert commit (9e20fb9) on a `TT-9424-` prefixed branch.

PR #1091, opened 2026-08-25
[TT-14423] stop the hybrid pump orphaning gorpc clients

pvormste:
## Problem

`HybridPump.connectRPC()` replaced `p.clientSingleton` with a freshly built `gorpc.Client` and
never stopped the one it was replacing.

A `gorpc.Client` runs `rpc_pool_size` (default 5) `clientHandler` goroutines, each an infinite
redial loop that exits **only** when `Client.Stop()` is called. Dropping the reference does
nothing. MDCB has no idle timeout and no connection cap, so it accepts every one of those redials
and pays roughly 1.7 MB and three goroutines per connection, permanently.

Every recovery therefore cost a whole pool that neither side ever released:

```
WriteData
  ├─ no records? return                          ← an idle pump never leaks
  ├─ RPCLogin() fails
  │    └─ ErrRPCLogin (bad credentials)? return  ← does not leak
  └─ any other error → connectAndLogin(false)
       └─ connectRPC()
            ├─ p.clientSingleton = gorpc.NewTLSClient(...)   ← old client abandoned
            └─ p.clientSingleton.Start()                     ← +5 goroutines, +5 connections
```

That closes a loop with no ceiling: MDCB runs out of memory → the restart outlasts the pump's
`call_timeout` → every pump orphans a client → MDCB runs out sooner.

Two conditions are needed to trigger it, and both are easy to get wrong. Records have to be
flowing, because `WriteData` returns early with nothing to write; and the failure has to outlast
`call_timeout` (default 10s), because gorpc redials within about a second, so merely severing
established connections leaks nothing.

`Init()` had the same defect and could orphan **three** clients on its own: it calls
`connectAndLogin(true)`, whose `retryAndLog` wrapper invokes `connectRPC` up to four times, and
each invocation built and started another client.

Present since v1.8.0 (`ca22ae4b`, March 2023).

### `connectRPC` releases the client it replaces

The new client is configured in a local variable and published only after `Start()` has returned.
That establishes the invariant the rest of the change relies on: **`clientSingleton` is non-nil
only when it refers to a running client**, so a non-nil pointer is always safe to stop exactly
once. The invariant is necessary because `gorpc.Client.Stop()` panics both on a client that was
never started and on one that is stopped twice, and `clientStopChan` is unexported — there is no
way to ask a client whether it is running.

The replaced client is stopped *before* the successor starts, in a new `stopClient()` helper that
also clears the pointers and resets `clientIsConnected`. Stopping first rather than last is
deliberate:

- `Stop()` returns promptly even when MDCB is unreachable — the pooled handlers return as soon as
  the stop channel closes and do not wait for an outstanding dial — so the gap is negligible.
- It keeps the pump within `rpc_pool_size` connections at every instant, including while a
  retrying connect works through its attempts. Starting first would spike to twice the pool per
  attempt, which is the shape of the incident this fixes.
- Only one client's `OnConnect` is ever live, so `clientIsConnected` has a single writer.

There is no healthy traffic to protect during that gap: `connectRPC` is only reached because a
`Ping` or `Login` has already failed, or from `Init`.

### The reconnect is serialised

`WriteData` runs on a goroutine that is abandoned when the pump's configured `timeout` expires,
and the next purge cycle starts another one, so two recoveries can be in flight at once. A new
`connectMu` is taken for the whole of `connectAndLogin`, covering the retry loop *and* the login
that follows. Locking only around a single connect attempt would let attempts interleave, each
stopping the client the other had just published.

Without it the damage was worse than a leak. `startDispatcher` used to assign
`p.dispatcher = gorpc.NewDispatcher()` and then register the functions into the field, so two
concurrent reconnects registered into whichever dispatcher won the race — a concurrent map write,
which crashes the process:

```
fatal error: concurrent map writes
  gorpc.(*Dispatcher).AddFunc
  pumps.(*HybridPump).startDispatcher
  pumps.(*HybridPump).connectRPC
```

`startDispatcher` now builds the dispatcher and func client in locals and publishes them as a
unit, which removes that crash independently of the lock.

A second, separate `clientMu sync.RWMutex` guards the two pointers, which `callRPCFn` reads from
the `WriteData` goroutines. It is deliberately not the same mutex: `connectRPC` ends by calling
`callRPCFn("Ping", ...)`, so a single non-reentrant lock would self-deadlock. Lock order is
`connectMu` then `clientMu`, never the reverse, and `clientMu` is only ever held to assign or read
pointers, never across an RPC — so an in-flight purge is never blocked behind a 10s `Ping`.

`callRPCFn` now returns an error when there is no client instead of dereferencing nil. That also
avoids burning a full `CallTimeout` on a client that has been stopped.

### Two adjacent defects, included because they are the same bug

- **`Shutdown()` did an unguarded `Stop()`.** `Init` returns before `connectRPC` for an invalid
  configuration, leaving `clientSingleton` nil, and a pump whose `Init` failed is still shut down
  along with the others — a live nil-pointer dereference. It now goes through `stopClient()` under
  `connectMu`, which is nil-safe and repeatable. The behaviour the existing
  `TestHybridPumpShutdown` asserts is unchanged.
- **`Init()` left a running client behind when only the login failed.** Nothing shuts down a pump
  whose `Init` returned an error, so a wrong `api_key` orphaned a full pool for the lifetime of the
  process — an independent leak path. `Init` now releases the client on that branch.

### Deliberately not changed

- `connectRPC` still returns the `Ping` error with the new client assigned. That is now correct and
  bounded: it is the pump's current client, exactly one pool, and every exit path releases it — the
  next `connectRPC`, `Shutdown()`, or the new `Init` cleanup. Keeping it assigned preserves how the
  pump recovers when MDCB comes back.
- The `OnConnect` method value bound to the pump is now moot rather than a hazard: `Stop()` waits
  for the handler goroutines, and `OnConnect` is only reachable from them, so no stopped client can
  write `clientIsConnected` afterwards.
- ~~A "someone else already reconnected, reuse their client" short-circuit.~~ Originally left out
  as an optimisation. Review showed it is not optional — see the addendum at the end, which adds
  it.

## Tests

Six new tests in `pumps/hybrid_test.go`, with two more and one extended assertion added by the
addendum, for eight in total. **Every one of them was confirmed to fail against the code it
guards**, which matters here because nothing about the leak is observable from the pump's own
behaviour.

| Test | What it pins down | Unpatched result |
|---|---|---|
| `TestConnectRPCStopsPreviousClient` | reconnecting leaves exactly one live client, and the replaced one is actually stopped | 2 clients × 3 connections, and `Stop()` on the old one does not panic |
| `TestConnectAndLoginRetryDoesNotOrphanClients` | the `retryAndLog` path builds several clients but leaves one | 3 clients × 2 connections |
| `TestConcurrentReconnectKeepsSingleClient` | overlapping recoveries leave one client, none stopped from under another, no torn state | 9 clients × 2 connections, or `fatal error: concurrent map writes` |
| `TestShutdownWithoutClient` | `Shutdown` on a pump that never connected | nil-pointer dereference |
| `TestShutdownIsIdempotent` | `Shutdown` twice | nil-pointer dereference |
| `TestInitFailureReleasesRPCClient` | a failed `Init` keeps no client | client kept, 2 connections open |

### How they observe the leak

The pump generates one connection id per `connectRPC` and writes it in its dial handshake, so
"connections still open for id *k*" is exactly "client *k* is still alive". The tests add a
`gorpc.Listener` that counts open connections per id, which is the only place the leak is visible.

Three constraints shaped that choice:

- **No `runtime.NumGoroutine()`.** The package's pre-existing tests already leak clients, and other
  pumps in it run their own background goroutines, so absolute or delta goroutine counts would be
  flaky and would misattribute failures.
- **No reliance on `-race`.** `bin/ci-test.sh` disables the race detector for the `pumps` package,
  so every assertion is a value assertion. `TestConcurrentReconnectKeepsSingleClient` fails
  deterministically without it: `connectRPC` only returns nil after its `Ping` succeeded, so
  pre-fix every one of the nine clients provably has a live connection.
- **No new fixed ports.** The listener owns an ephemeral socket and its `Init` is a no-op, so these
  tests cannot collide with the file's hardcoded 12345/9092/9099 across the three times CI runs this
  package.

`TestConnectRPCStopsPreviousClient` also asserts that stopping the replaced client panics. That is
the only *direct* evidence available that `connectRPC` stopped it, as opposed to its connections
having gone away for some other reason. It is asserted last, because it stops the client itself.

One measured gorpc behaviour is documented in the tests via `rpcReleaseBudget`: after
`Client.Stop()` returns, the server closes its end of each connection either immediately or after
almost exactly ten seconds, never later. This reproduces with no pump involved at all, so the
waits have to clear those ten seconds. An unpatched pump never releases the connections at all, so
the longer budget does not weaken any assertion.

### One drive-by in the test file

`TestHybridConfigParsing` unset its environment variables **by value instead of by key**
(`for _, env := range envs { os.Unsetenv(env) }`, where `env` is `"localhost:9099"`, `"20"`, …),
so its `TYK_PMP_PUMPS_HYBRID_META_*` settings stayed set for every test declared after it — which
is where new tests land. Now iterates keys.

## Verification

Unit tests aside, the fix was checked end to end on a local data plane (worker gateway → worker
Redis → hybrid pump → MDCB) with records flowing continuously and 20 interruptions of the
pump→MDCB path, each longer than `call_timeout`:

| | before | after |
|---|---|---|
| reconnects | 20 | 21 |
| pump established connections | **105** (21 clients × 5) | **5** |
| pump `clientHandler` goroutines | **105** | **5** |
| MDCB connections | 110 | **10** |
| MDCB goroutines (start → peak) | 54 → **339** | 54 → **54** |
| MDCB heap in use | 28 MB → **203 MB** | 27 MB → 36 MB |
| pump goroutines (peak) | 306 | 23 |

Before the fix the growth was exactly linear — connections were always `(reconnects + 1) × 5` and
never dropped. After it they stay at `rpc_pool_size` however often the pump reconnects, and MDCB
returns to its pre-test baseline instead of settling higher.

`gofmt -s`, `goimports`, `go vet` and `golangci-lint` are clean. The `pumps` package passes with
both `MONGO_DRIVER` values, under `-tags=boringcrypto`, and additionally under `-race` (which CI
does not run for this package).

## Scope

tyk-pump only. Two related items are deliberately out of scope and tracked separately: MDCB-side
`server_options.idle_timeout` / `max_connections` hardening, which bounds the damage from any
misbehaving client but does not fix this leak; and a distinct race in the gateway's
`rpc/dns_resolver.go`, which is newer than the reported incidents and cannot explain them.

The defect predates every supported release, so this needs backporting to the supported release
lines.

---

# Addendum: reconnect stampede could break a purge in flight

Raised in review, and valid. Worth stating plainly: **this was a regression introduced by the fix
above, and it loses analytics data.** It is fixed here, in the same change.

## What happens

Several purge cycles can be in flight at once. When the pump has a `timeout` configured,
`execPumpWriting` returns as soon as that deadline passes while the goroutine it started keeps
running `WriteData`; the wait group is satisfied, the purge loop moves on, and the next tick starts
another `WriteData` on the same pump.

If MDCB has gone away, all of those calls fail `RPCLogin` together and all call
`connectAndLogin(false)`, where they queue on `connectMu`. Before this addendum each of them then
reconnected unconditionally:

1. A takes the lock, reconnects, publishes client A, releases the lock, returns nil.
2. A goes on to `callRPCFn("PurgeAnalyticsData")` through client A.
3. B takes the lock and calls `connectRPC()`, which calls `stopClient()`.
4. B stops client A underneath A.
5. A's call fails — or finds `funcClientSingleton` already nil and gets `not connected to Tyk MDCB`.

**Those records are then gone.** `StartPurgeLoop` reads analytics with
`AnalyticsStore.GetAndDeleteSet(...)`, so they have already been deleted from the temporal store by
the time any pump sees them, and `execPumpWriting` reports a failed write as
`Warning("Error Writing to: …")` with no retry and no re-queue.

The reason this is new: before the fix, the replaced client was never stopped. It leaked, but it
stayed connected and usable, so a call in flight through it still completed. Adding `stopClient()`
turned a silent leak into a broken write. Only step 4 is new behaviour, and it is the step that
costs data.

## The fix

A generation counter, `clientGen atomic.Uint64`, incremented wherever a client is published.
`connectAndLogin` reads it **before** queueing on the lock, and once it holds the lock a changed
value means another recovery has already replaced the client this one set out to replace. It then
probes that client with `RPCLogin` and, if it answers, returns without touching it.

Reading the generation before taking the lock rather than after is the part that matters. After
would still rebuild in the case where the failure was observed on the old client but the read
happened once the replacement was already published.

Rejected credentials are returned rather than reconnected around: `ErrRPCLogin` means another client
would be given the same answer, so rebuilding would only tear down a working connection.

Two useful consequences beyond the data loss:

- The stampede collapses. Eight overlapping recoveries now build **one** client between them
  instead of eight, so a reconnect no longer multiplies into a burst of connects against an MDCB
  that has only just come back — which is the condition this whole ticket is about.
- The common path is unaffected. A lone recovery sees an unchanged generation and goes straight to
  reconnecting, with no extra round trip.

## Tests

Three, all confirmed to fail against the fix *without* the short-circuit — that is, against the
state of this branch before the addendum, which is the code they actually guard.

| Test | Result without the short-circuit |
|---|---|
| `TestQueuedRecoveryReusesTheNewClient` | the queued recovery built another client and replaced the working one; the failure output shows the published client with `clientStopChan: nil`, stopped while a purge was using it |
| `TestQueuedRecoveryKeepsClientWhenCredentialsAreRejected` | a rejected login triggered a rebuild and replaced a working client |
| `TestConcurrentReconnectKeepsSingleClient` (extended) | **9** clients built, against a bound of 3 |

The first two encode the numbered sequence above deterministically rather than racing for it: the
test file is in the same package, so it holds `connectMu` itself, lets a second recovery queue
behind it, publishes a client under the lock, and then releases. `TestQueuedRecoveryReusesTheNewClient`
finishes by making a real `PurgeAnalyticsData` call through the reused client, so the assertion is
the write path itself rather than a proxy for it.

## One residual, deliberately

A *single* recovery that legitimately reconnects — its own login genuinely failed — while another
purge is mid-write will still stop that client, and that write still fails. Closing it needs
per-client reference counting so `stopClient` drains in-flight calls first, which means either
holding a lock across an RPC or wrapping every client.

Not done here, on the grounds that the client is broken by hypothesis in that case, so the write was
already unlikely to complete, and before this change it only appeared to succeed because it was
riding a leaked client.

Related but wider than this PR: **the data loss itself is not hybrid-specific.** Any pump loses its
batch when `WriteData` fails, because the purge loop deletes the records before handing them over
and a failure is only logged. That is worth its own ticket.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-14423" title="TT-14423" target="_blank">TT-14423</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | High memory consumption on MDCB caused by leaked RPC clients in the hybrid Pump |

Generated at: 2026-08-26 10:12:48

</details>

### DIFF
--- a/pumps/hybrid.go
+++ b/pumps/hybrid.go
@@ -68,6 +68,10 @@ type HybridPump struct {
 	// WriteData goroutines while a reconnect may be replacing them.
 	clientMu sync.RWMutex
 
+	// clientGen counts published clients, so a recovery that waited for connectMu can tell
+	// whether the client whose failure sent it there has already been replaced.
+	clientGen atomic.Uint64
+
 	clientSingleton   *gorpc.Client
 	dispatcher        *gorpc.Dispatcher
 	clientIsConnected atomic.Value
@@ -201,6 +205,8 @@ func (p *HybridPump) startDispatcher(client *gorpc.Client) {
 	p.dispatcher = dispatcher
 	p.clientSingleton = client
 	p.funcClientSingleton = funcClient
+
+	p.clientGen.Add(1)
 }
 
 // stopClient stops the current RPC client, if there is one, and clears it.
@@ -472,11 +478,35 @@ func (p *HybridPump) sendMCPAggregates(data []interface{}) error {
 
 // connectAndLogin connects to RPC server and logs in if retry is true, it will retry with retryAndLog func
 func (p *HybridPump) connectAndLogin(retry bool) error {
+	// Which client this recovery is trying to replace. Read before queueing on the lock, so it
+	// can be compared with whatever is current once the lock is held.
+	observedGen := p.clientGen.Load()
+
 	// One recovery at a time, covering the retry loop and the login that follows it. Locking
 	// only around a single connect attempt would let two recoveries interleave, each stopping
 	// the client the other had just published.
 	p.connectMu.Lock()
 	defer p.connectMu.Unlock()
+
+	// When several purge cycles overlap they all fail together and all end up here, queued on
+	// the lock. Only the first of them needs to reconnect: rebuilding again would stop a client
+	// that is already working and that another purge may be part way through writing through,
+	// and a purge that fails loses its records for good, because the purge loop deletes them
+	// from the temporal store before handing them over.
+	if p.clientGen.Load() != observedGen {
+		err := p.RPCLogin()
+		if err == nil {
+			p.log.Debug("Reusing the MDCB connection another purge just established")
+
+			return nil
+		}
+
+		// Credentials are wrong rather than the connection being broken. Reconnecting cannot
+		// fix that, and doing so would tear down a usable client for nothing.
+		if errors.Is(err, ErrRPCLogin) {
+			return err
+		}
+	}
 
 	connectFn := p.connectRPC
 	loginFn := p.RPCLogin
--- a/pumps/hybrid.go
+++ b/pumps/hybrid.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"net"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -53,6 +54,19 @@ var (
 // HybridPump allows to send analytics to MDCB over RPC
 type HybridPump struct {
 	CommonPumpConfig
+
+	// connectMu serialises connecting, reconnecting and shutting down. A WriteData call that
+	// outlives the configured pump timeout is abandoned and started again on the next purge
+	// cycle, so two recoveries can be in flight at once; without this lock each would build a
+	// client and could stop the other's brand-new one.
+	//
+	// Lock order is connectMu then clientMu, never the reverse. clientMu is only ever held to
+	// assign or read the pointers below, never across an RPC.
+	connectMu sync.Mutex
+
+	// clientMu guards clientSingleton and funcClientSingleton, which callRPCFn reads from the
+	// WriteData goroutines while a reconnect may be replacing them.
+	clientMu sync.RWMutex
 
 	clientSingleton   *gorpc.Client
 	dispatcher        *gorpc.Dispatcher
@@ -149,22 +163,74 @@ func (p *HybridPump) Init(config interface{}) error {
 
 	if err := p.connectAndLogin(true); err != nil {
 		p.log.Error(err)
+
+		// The connect may well have succeeded and only the login failed, leaving a running
+		// client behind. Nothing shuts down a pump whose Init returned an error, so its pool
+		// would stay up for the lifetime of the process.
+		p.connectMu.Lock()
+		p.stopClient()
+		p.connectMu.Unlock()
+
 		return err
 	}
 
 	return nil
 }
 
-func (p *HybridPump) startDispatcher() {
-	p.dispatcher = gorpc.NewDispatcher()
+// startDispatcher publishes an already started client along with a dispatcher client bound to
+// it. A gorpc.DispatcherClient cannot be pointed at a different gorpc.Client, so both have to be
+// rebuilt for every connection.
+//
+// The dispatcher is built in a local and published as a whole. Assigning it to the pump first
+// and filling it in afterwards let two concurrent reconnects register their functions into
+// whichever dispatcher happened to win the field, which is a concurrent map write.
+//
+// The caller must hold connectMu.
+func (p *HybridPump) startDispatcher(client *gorpc.Client) {
+	dispatcher := gorpc.NewDispatcher()
 
 	for funcName, funcBody := range dispatcherFuncs {
-		p.dispatcher.AddFunc(funcName, funcBody)
+		dispatcher.AddFunc(funcName, funcBody)
 	}
 
-	p.funcClientSingleton = p.dispatcher.NewFuncClient(p.clientSingleton)
+	funcClient := dispatcher.NewFuncClient(client)
+
+	p.clientMu.Lock()
+	defer p.clientMu.Unlock()
+
+	p.dispatcher = dispatcher
+	p.clientSingleton = client
+	p.funcClientSingleton = funcClient
 }
 
+// stopClient stops the current RPC client, if there is one, and clears it.
+//
+// A client is published only after gorpc.Client.Start() has returned (see connectRPC), so a
+// non-nil clientSingleton is always a running client and is safe to stop exactly once - gorpc
+// panics both on stopping a client that was never started and on stopping one twice.
+//
+// The caller must hold connectMu.
+func (p *HybridPump) stopClient() {
+	p.clientMu.Lock()
+	client := p.clientSingleton
+	p.clientSingleton = nil
+	p.funcClientSingleton = nil
+	p.clientMu.Unlock()
+
+	p.clientIsConnected.Store(false)
+
+	if client == nil {
+		return
+	}
+
+	// Bounded even when MDCB is unreachable: the pooled handlers return as soon as the client's
+	// stop channel is closed and do not wait for an outstanding dial.
+	client.Stop()
+}
+
+// connectRPC replaces the current RPC client with a freshly connected one.
+//
+// The caller must hold connectMu.
 func (p *HybridPump) connectRPC() error {
 	p.log.Debug("Setting new MDCB connection!")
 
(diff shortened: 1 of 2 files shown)


pumps/hybrid.go
```diff
@@ -381,6 +472,12 @@ func (p *HybridPump) sendMCPAggregates(data []interface{}) error {
 
 // connectAndLogin connects to RPC server and logs in if retry is true, it will retry with retryAndLog func
 func (p *HybridPump) connectAndLogin(retry bool) error {
+	// One recovery at a time, covering the retry loop and the login that follows it. Locking
```
vladzabolotnyi, line 485: I've pinged LLM to find concurrency edge cases that can affect current implementation and there is one bothering me. I can't validate it fully but it looks valid from my POV.
```
Thundering Herd causing Data Loss
The Problem:
In pumps/hybrid.go, connectAndLogin unconditionally calls connectRPC() once it acquires connectMu.Lock().
If the connection to MDCB drops, multiple concurrent WriteData calls (from different purge cycles) will fail RPCLogin and call connectAndLogin(false). They will all queue up on connectMu.Lock().

Here is the sequence of events:
1. Goroutine A acquires the lock, successfully reconnects (creating Client A), unlocks, and returns nil.
2. Goroutine A proceeds to call callRPCFn("PurgeAnalyticsData") using Client A.
3. Goroutine B acquires the lock and unconditionally calls connectRPC(), which calls stopClient().
4. Goroutine B stops Client A while Goroutine A is trying to use it.
5. Goroutine A's RPC call fails.

Because Tyk Pump pops analytics data from Redis before calling WriteData, and WriteData failures are only logged as warnings, Goroutine A's analytics data is permanently lost.
```

vladzabolotnyi (changes requested):

pumps/hybrid.go
pvormste: confirmed, and fixed with latest commit

vladzabolotnyi (approved):
🚀

